PR #60, opened 2023-05-19
Add py.typed to dbt_semantic_interfaces

tlento:
We use typehints, but without publishing that fact the package is not
PEP-561 compliant, and as such mypy will not be able to follow imports
and resolve typehints.

This adds the py.typed to the package so that we support typehinting
from the moment of our initial deployment.

resolves #

### Checklist


QMalcolm (approved):
Good catch! Yea we absolutely need this

PR #66, opened 2023-05-23
Use Caching in Github Actions Workflow

plypaul:
### Description

This PR sets up the caching action in the Github workflow so that installing Python dependencies is faster.
### Checklist

### DIFF
--- a/.github/actions/setup-python-env/action.yml
+++ b/.github/actions/setup-python-env/action.yml
@@ -4,15 +4,42 @@ inputs:
   python-version:
     description: Python version number to install
     required: true
+  additional-cache-key:
+    description: |
+      An additional cache key that can be specified to differentiate workflows that install a different set of packages.
+    required: false
+    default: "default"
 
 runs:
   using: "composite"
   steps:
+  - name: Set Linux Release Environment Variable
+    shell: bash
+    run: echo "LINUX_RELEASE=$(lsb_release -rs)" >> $GITHUB_ENV
   - name: Set up Python ${{ inputs.python-version }}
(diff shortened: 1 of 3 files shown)


QMalcolm (approved):
Looks good to me! Excited to see how much this speeds things up 🚀

.github/actions/setup-python-env/action.yml
```diff
@@ -1,14 +1,51 @@
...
+name: Setup Python / Hatch / Pydantic Environment
+description: Install Python / Hatch / Pydantic
+
+inputs:
+  python-version:
+    description: "Version of Python to use for testing"
+    required: true
+  pydantic-version:
+    description: "Version of Pydantic to use for testing"
+    required: true
```
tlento, line 10: If you set a default of 1.10 on this I think you can revert most of the non-caching updates in this PR, since I've separately consolidated actions into this setup.

Alternatively, we can use this as an annotation defaulted to "none" just to seed the cache key and then only install Pydantic in the calling action where it's required. That'll eliminate nearly all of the action-level changes in this PR. If we go that route we should probably name the key something like `installed-pydantic-version` or suchlike.

tlento (commented):
I like the caching but in updating our actions to cover Python 3.12 I have clobbered a bunch of your changes here. I've left a comment inline about how to make the conflict resolution a bit simpler.

plypaul:
My delays in merging approved PRs have come to bite me.

plypaul:
FYI - this reduces the Python test action from 90-120s -> 30-40s.

PR #72, opened 2023-05-25
Use Common Hatch Environment for Testing & Linting + Parallelize Tests

plypaul:
### Description

This PR updates `pyproject.toml` so that it uses a single `hatch` environment (called `dev-env`) for testing and linting. The benefit of using a single environment is that it simplifies the setup, and there doesn't seem to be much of a benefit for having separate dependencies for those tasks.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,13 +8,13 @@ install-hatch:
 
 # This edits your local pre-commit hook file to use Hatch when executing.
 overwrite-pre-commit:
-	hatch run pre-commit:pre-commit install
-	hatch run pre-commit:sed -i -e "s/exec /exec hatch run pre-commit:/g" .git/hooks/pre-commit
+	hatch run dev-env:pre-commit install
+	hatch run dev-env:sed -i -e "s/exec /exec hatch run dev-env:/g" .git/hooks/pre-commit
 
 install: install-hatch overwrite-pre-commit
 
 test:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -8,13 +8,13 @@ install-hatch:
 
 # This edits your local pre-commit hook file to use Hatch when executing.
 overwrite-pre-commit:
-	hatch run pre-commit:pre-commit install
-	hatch run pre-commit:sed -i -e "s/exec /exec hatch run pre-commit:/g" .git/hooks/pre-commit
+	hatch run dev-env:pre-commit install
+	hatch run dev-env:sed -i -e "s/exec /exec hatch run pre-commit:/g" .git/hooks/pre-commit
```
QMalcolm, line 12: I pulled down the branch and made an empty commit just to do a sanity check. Doing so gave me the error
`Unknown environment: pre-commit`. It looks like the follow change needs to be made so that hatch runs pre-commit properly.
```suggestion
	hatch run dev-env:sed -i -e "s/exec /exec hatch run dev-env:/g" .git/hooks/pre-commit
```

QMalcolm (approved):
This looks great! We do need to make one small change to correct hatch command for running pre-commit

Makefile
plypaul: Makes sense - thanks!

PR #73, opened 2023-05-25
Protocol Tests

plypaul:
### Description

A few tests using protocols to evaluate the signature change.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/protocol_example.py
+++ b/dbt_semantic_interfaces/protocol_example.py
@@ -1,0 +1,107 @@
+import logging
+from abc import abstractmethod
+from dataclasses import dataclass
+from typing import Protocol, Sequence
+
+from dbt_semantic_interfaces.implementations.base import FrozenBaseModel
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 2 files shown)


dbt_semantic_interfaces/protocol_example.py
```diff
@@ -0,0 +1,107 @@
...
+#
+
+
+# Implicit Pydantic implementation
+class ImplicitPydanticSemanticModel(FrozenBaseModel):  # noqa: D
+    name: str
+
+
+class ImplicitPydanticSemanticManifest(FrozenBaseModel):  # noqa: D
+    semantic_models: Sequence[ImplicitPydanticSemanticModel]
```
tlento, line 46: Interesting. I think this works because the protocol only defines the property accessor, not the property setter, so this can be effectively added as a covariant protocol type. That seems like exactly what we need. Assuming setting the properties with ImplicitPydanticSemanticModel subclasses works as expected I think we'll be good to go here.

tlento (commented):
Looks like the right way to do it to me.

PR #74, opened 2023-05-30
Change Definition Protocols to Use An Abstract Property

plypaul:
### Description

This PR changes the protocols used for defining the semantic manifest to use an abstract property as it allows us to use the existing Pydantic implementations.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/elements/dimension.py
+++ b/dbt_semantic_interfaces/implementations/elements/dimension.py
@@ -17,7 +17,7 @@ from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
 ISO8601_FMT = "YYYY-MM-DD"
 
 
-class DimensionValidityParams(HashableBaseModel):
+class PydanticDimensionValidityParams(HashableBaseModel):
     """Parameters identifying a given dimension as an entity for validity state.
 
     This construct is used for supporting SCD Type II tables, such as might be
@@ -31,22 +31,22 @@ class DimensionValidityParams(HashableBaseModel):
     is_end: bool = False
 
 
-class DimensionTypeParams(HashableBaseModel):
-    """Dimension type params add additional context to some types (time) of dimensions."""
+class PydanticDimensionTypeParams(HashableBaseModel):
+    """PydanticDimension type params add additional context to some types (time) of dimensions."""
 
     is_primary: bool = False
     time_granularity: TimeGranularity
-    validity_params: Optional[DimensionValidityParams] = None
+    validity_params: Optional[PydanticDimensionValidityParams] = None
 
 
-class Dimension(HashableBaseModel, ModelWithMetadataParsing):
+class PydanticDimension(HashableBaseModel, ModelWithMetadataParsing):
     """Describes a dimension."""
 
     name: str
     description: Optional[str]
     type: DimensionType
     is_partition: bool = False
-    type_params: Optional[DimensionTypeParams]
+    type_params: Optional[PydanticDimensionTypeParams]
     expr: Optional[str] = None
     metadata: Optional[Metadata]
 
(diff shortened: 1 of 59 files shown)


dbt_semantic_interfaces/implementations/elements/dimension.py
```diff
@@ -17,7 +17,7 @@
 ISO8601_FMT = "YYYY-MM-DD"
 
 
-class DimensionValidityParams(HashableBaseModel):
+class PydanticDimensionValidityParams(HashableBaseModel):
```
tlento, line 20: The `Pyantic<thing>` nomenclature bugs me, because the Pydantic-ness of it seems like an internal implementation detail around how we deal with parsing and serialization rather than something we should broadcast throughout the codebase. Maybe we call these `MetricFlow<thing>` or something to indicate that these are the default implementations that MetricFlow expects? 

On the other hand, if we're going to house different serialization mechanisms in this repo, and that's the key differentiator on the implementations, then this naming approach is excellent.

tests/validations/test_unique_valid_name.py
```diff
@@ -19,11 +19,11 @@
     Top level elements include
     - Semantic Models
     - Metrics
-    - Metric Sets
+    - PydanticMetric Sets
     - Dimension Sets
 
     A name for any of these elements must be unique to all other top level elements
-    except metrics. Metric names only need to be unique in comparison to other metric
+    except metrics. PydanticMetric names only need to be unique in comparison to other metric
```
tlento, line 28: Neither of these need to be changed, the Metric here is the generic construct rather than the implementation class.

dbt_semantic_interfaces/validations/element_const.py
```diff
@@ -51,7 +53,7 @@ def validate_model(model: SemanticManifest) -> Sequence[ValidationIssue]:  # noq
 
     @staticmethod
     def _get_element_name_to_types(
-        model: SemanticManifest,
+        model: PydanticSemanticManifest,
```
tlento, line 56: I got this far before it occurred to me, but wow, we use `model` in a lot of places that should now be `manifest` or `semantic_manifest`

This is a problem for later.

dbt_semantic_interfaces/test_utils.py
```diff
@@ -38,8 +38,8 @@ def as_datetime(date_string: str) -> datetime.datetime:
 
 
 def find_semantic_model_with(
-    model: PydanticSemanticManifest, function: Callable[[SemanticModel], bool]
-) -> Tuple[SemanticModel, int]:
+    model: PydanticSemanticManifest, function: Callable[[PydanticSemanticModel], bool]
+) -> Tuple[PydanticSemanticModel, int]:
```
tlento, line 42: I think this is the right thing to do for now but I suspect we should eventually move most of our test shims to use the protocols. We can open an issue or something for this later.

dbt_semantic_interfaces/validations/common_entities.py
```diff
@@ -67,7 +65,7 @@ def _check_entity(
 
     @staticmethod
     @validate_safely(whats_being_done="running model validation warning if entities are only one one semantic model")
-    def validate_model(model: PydanticSemanticManifest) -> Sequence[ValidationIssue]:
+    def validate_model(model: SemanticManifest) -> Sequence[ValidationIssue]:
```
tlento: Little unexpected here, but I like it.

tlento (approved):
Sweet, thanks for breaking this up so cleanly, too, made for a much easier read!

Some alternatives to `Pydantic` as the name qualifier:

- Basic
- Stock
- Standard
- MetricFlow
- MF

If we decide to change the naming I suspect it's way easier to just do global symbol renames at some later point in time so let's just get this in. I don't think any of these are truly public-facing anyway so I don't think there's any hurry.

tests/validations/test_unique_valid_name.py
plypaul: Let me fix this at the end of the stack.

plypaul:
Up to rename `Pydantic...` to something else, but seeing what the best proposal is. Also something that can be handled fairly easily now that we have a unique-ish name in place.

PR #75, opened 2023-05-30
Change Validations to be Generic

### LINKED ISSUE #20: Update `SemanticManifestValidator` to expect a generic bound by the `SemanticManifest` Protocol
Currently the `SemanticManifestValidator` (currently named `ModelValidator`) expects a `UserConfiguredModel` which is a concrete object. Initially we thought we should move to it expecting `SemanticManifest` protocol. However, in dbt-core we'll be writing nodes which extend the protocol definition. If we want to be able to write validation rules that can operate on the extensions and guarantee type safety, then we need to take it a step further. Thus the `SemanticManifestValidator` should instead operate on a generic bound by the `SemanticManifest` protocol.

Something like...

``` Python
from typing import TypeVar
from dbt_semantic_interfaces.protocols import SemanticManifest


T = TypeVar("T", bound="SemanticManifest")
class SemanticManifestValidator:
  ...

  def validate(self, semantic_manifest: T) -> ValidationResults:
    ...
```

plypaul:
Resolves #20 

### Description

This PR changes the signature of the classes used for validation to be generic so that other implementations of the `SemanticManifest` can use the same framework with type-safety.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/validator_helpers.py
+++ b/dbt_semantic_interfaces/validations/validator_helpers.py
@@ -12,10 +12,8 @@ import click
 from pydantic import BaseModel, Extra
 
 from dbt_semantic_interfaces.implementations.base import FrozenBaseModel
-from dbt_semantic_interfaces.implementations.semantic_manifest import (
-    PydanticSemanticManifest,
-)
 from dbt_semantic_interfaces.protocols.metadata import Metadata
+from dbt_semantic_interfaces.protocols.semantic_manifest import SemanticManifest
 from dbt_semantic_interfaces.references import (
     MetricModelReference,
     SemanticModelElementReference,
@@ -383,7 +381,7 @@ class ModelValidationRule(ABC):
 
     @classmethod
     @abstractmethod
-    def validate_model(cls, model: PydanticSemanticManifest) -> Sequence[ValidationIssue]:
+    def validate_model(cls, model: SemanticManifest) -> Sequence[ValidationIssue]:
         """Check the given model and return a list of validation issues."""
         pass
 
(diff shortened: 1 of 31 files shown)


dbt_semantic_interfaces/validations/validator_helpers.py
```diff
@@ -385,18 +385,6 @@ def validate_model(cls, model: SemanticManifest) -> Sequence[ValidationIssue]:
...
-    def validate_model_serialized_for_multiprocessing(cls, serialized_model: str) -> str:
-        """Validate a model serialized via Pydantic's .json() method, and return a list of JSON serialized issues.
-
-        This method exists because our validations are forked into parallel processes via
-        multiprocessing.ProcessPoolExecutor, and passing a model or validation results object can result in
-        idiosyncratic behavior and inscrutable errors due to interactions between pickling and pydantic objects.
-        """
-        return SemanticManifestValidationResults.from_issues_sequence(
-            cls.validate_model(SemanticManifest.parse_raw(serialized_model))
-        ).json()
```
tlento: If I remember correctly this sometimes but not always threw errors. Like I had persistent failures in CI but not locally or vice-versa when this popped up. It's possible pydantic fixed something, or that we cut out the bits that were causing problems (like the special input handing in our old WhereClauseConstraint objects), but it's also possible we just haven't stumbled in a bad configuration yet.

I think it's fine to remove it here, especially given the scope of this PR, but I wouldn't be surprised if we had to put it back. The best practice here is to pass serialized data across the process boundary rather than passing data objects and relying on pickling's best-effort serialization. There's a lot to be said for the key protocols having `serialize()` and `deserialize()` or similar methods, which would allow us to avoid the pickling as well.

tlento (approved):
Nice, thanks! The pickling dependency could be an issue at some point but we have options for dealing with it if it causes problems, so let us proceed.

dbt_semantic_interfaces/validations/validator_helpers.py
plypaul: I got a reproducible failure case, and the issue actually was due to the way that Pydantic handles pickling if there is a dataclass embedded within a Pydantic object. Pickling of the definition Pydantic objects is fine, but pickling of `SemanticManifestValidationResults` is not, so I changed it to only JSON-serialize that class.

PR #76, opened 2023-05-30
Remove `--ignore-missing-imports` from Mypy and Add Pydantic Plugin

plypaul:
### Description

This PR improves type checking in the project by removing the `--ignore-missing-imports` flag for the mypy pre-commit check and adding the Pydantic Mypy plugin. The plugin allows for more checks on Pydantic types.

### Checklist

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,8 @@
+[mypy]
+plugins = pydantic.mypy
+
+[pydantic-mypy]
+init_forbid_extra = True
+init_typed = True
+warn_required_dynamic_aliases = True
+warn_untyped_fields = True
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,6 +32,6 @@ repos:
     rev: v1.3.0
     hooks:
     - id: mypy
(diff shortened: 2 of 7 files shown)


mypy.ini
```diff
@@ -0,0 +1,8 @@
+[mypy]
+plugins = pydantic.mypy
+
+[pydantic-mypy]
+init_forbid_extra = True
```
tlento, line 5: We might need to remove this, but @QMalcolm can make that adjustment when it becomes an issue.

tlento (approved):
YES!

mypy.ini
QMalcolm: Is this the `forbid_extra` which makes it so that extras during pydantic deserialization are complained about? Or does this `forbid_extra` do something else?

plypaul: When `init_forbit_extra = True`, mypy throws an error for the following case:

```
class ExampleObject(FrozenBaseModel):  # noqa: D
    field0: int

def some_func() -> None:  # noqa: D
    logger.error(ExampleObject(field0=0, field1=2))
```

Otherwise, mypy does not.

PR #77, opened 2023-05-30
Change Transforms to be Generic

### LINKED ISSUE #21: Update SemanticManifestTransformer to expect a generic bound by the `SemanticManifest` Protocol
Currently the `SemanticManifestTransformer` (currently named `ModelTransformer`) expects and returns a `UserConfiguredModel` which is a concrete object. Initially we thought we should move to it expecting and returning a `SemanticManifest` protocol. However, in dbt-core we'll be writing nodes which extend the protocol definition. Additionally we want to be able to hand in raw-ish parsings and transform them into the final objects, meaning the input and return types should be different. If we want to be able to write transformation rules that can operate on the raw parsed objects, return the extended classes, and guarantee type safety, then we need to take it a step further. Thus the `SemanticManifestTransformer` should instead expect a generic and return a generic bound by the `SemanticManifest` protocol.

Something like...

``` Python
from typing import TypeVar
from dbt_semantic_interfaces.protocols import SemanticManifest


T = TypeVar("T", bound="SemanticManifest")
U = TypeVar("U")

class SemanticManifestTransformer:
  ...

  @staticmethod
  def transform(
    raw_semantic_manifest: U,
    ordered_rule_sequences: Tuple[Sequence[SemanticManifestTransformRule], ...] = DEFAULT_RULES,
  ) -> T:
    ....
```

plypaul:
Resolves #21 

### Description

This PR changes the signature of the classes used for semantic manifest transforms to be generic so that other implementations of the SemanticManifest can use the same transformation framework with type-safety.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/model_transformer.py
+++ b/dbt_semantic_interfaces/model_transformer.py
@@ -15,7 +15,7 @@ from dbt_semantic_interfaces.transformations.transform_rule import (
 logger = logging.getLogger(__name__)
 
 
-class SemanticManifestTransformer:
+class PydanticSemanticManifestTransformer:
     """Helps to make transformations to a model for convenience.
 
     Generally used to make it more convenient for the user to develop their model.
--- a/dbt_semantic_interfaces/model_transformer.py
+++ b/dbt_semantic_interfaces/model_transformer.py
@@ -20,23 +20,25 @@ from dbt_semantic_interfaces.transformations.convert_median import (
 )
 from dbt_semantic_interfaces.transformations.names import LowerCaseNamesRule
 from dbt_semantic_interfaces.transformations.proxy_measure import CreateProxyMeasureRule
-from dbt_semantic_interfaces.transformations.transform_rule import ModelTransformRule
+from dbt_semantic_interfaces.transformations.transform_rule import (
+    SemanticManifestTransformRule,
+)
 
 logger = logging.getLogger(__name__)
 
 
-class ModelTransformer:
+class SemanticManifestTransformer:
     """Helps to make transformations to a model for convenience.
 
     Generally used to make it more convenient for the user to develop their model.
     """
 
-    PRIMARY_RULES: Sequence[ModelTransformRule] = (
+    PRIMARY_RULES: Sequence[SemanticManifestTransformRule] = (
         LowerCaseNamesRule(),
         SetMeasureAggregationTimeDimensionRule(),
     )
 
-    SECONDARY_RULES: Sequence[ModelTransformRule] = (
+    SECONDARY_RULES: Sequence[SemanticManifestTransformRule] = (
         CreateProxyMeasureRule(),
         BooleanMeasureAggregationRule(),
         ConvertCountToSumRule(),
@@ -44,7 +46,7 @@ class ModelTransformer:
         AddInputMetricMeasuresRule(),
     )
 
-    DEFAULT_RULES: Tuple[Sequence[ModelTransformRule], ...] = (
+    DEFAULT_RULES: Tuple[Sequence[SemanticManifestTransformRule], ...] = (
         PRIMARY_RULES,
         SECONDARY_RULES,
     )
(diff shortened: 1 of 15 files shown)


dbt_semantic_interfaces/model_transformer.py
```diff
@@ -33,28 +21,10 @@ class SemanticManifestTransformer:
     Generally used to make it more convenient for the user to develop their model.
     """
 
-    PRIMARY_RULES: Sequence[SemanticManifestTransformRule] = (
```
tlento: It would be nice to have some indication of why this change is useful or necessary in the commit message.

dbt_semantic_interfaces/transformations/rule_set.py
```diff
@@ -0,0 +1,31 @@
...
+    @property
+    @abstractmethod
+    def secondary_rules(self) -> Sequence[SemanticManifestTransformRule]:
+        """TODO: Define what secondary means."""
+        pass
+
+    @property
+    @abstractmethod
+    def all_rules(self) -> Sequence[Sequence[SemanticManifestTransformRule]]:
+        """TODO: Why a nested sequence?"""
```
tlento, line 31: Oh yeah, this is a legacy holdover from when we used to do transform -> validate -> transform -> validate, except we never actually did that, it's just the interfaces all pretended like we did. 

I don't think we ever break up the transform rules so we can simply unnest and do primary + secondary here.

dbt_semantic_interfaces/transformations/rule_set.py
```diff
@@ -0,0 +1,31 @@
...
+    @abstractmethod
+    def primary_rules(self) -> Sequence[SemanticManifestTransformRule]:
+        """TODO: Define what primary means."""
+        pass
+
+    @property
+    @abstractmethod
+    def secondary_rules(self) -> Sequence[SemanticManifestTransformRule]:
+        """TODO: Define what secondary means."""
+        pass
```
tlento: Secondary rules run after all primary rules are complete. The execution of the rules can happen in arbitrary order within each level, but primary will always block secondary.

In practice I think we have primary "basic" transforms that make sense to apply before doing anything else.

dbt_semantic_interfaces/model_transformer.py
```diff
@@ -15,24 +17,36 @@
...
 
         It's important to note that some rules need to happen before or after other rules. Thus rules
         are passed in as an ordered tuple of rule sequences. Primary rules are run first, and then
         secondary rules. We don't currently have tertiary, quaternary, or etc currently, but this
         system easily allows for it.
         """
+        pass
+
+
+class PydanticSemanticManifestTransformer:
```
tlento, line 42: Should this declare the concrete type for the protocol generic? It seems like it should, or at least we'd want it to, but I'm not entirely clear on how concrete implementations on generic typed protocols work.

dbt_semantic_interfaces/model_transformer.py
```diff
@@ -15,24 +17,36 @@
...
+    """Transforms PydanticSemanticManifest."""
+
+    @staticmethod
+    def transform(  # noqa: D
+        model: PydanticSemanticManifest,
+        ordered_rule_sequences: Optional[Sequence[Sequence[SemanticManifestTransformRule]]] = None,
+    ) -> PydanticSemanticManifest:
```
tlento: This is all good enough for our purposes, so as long as we aren't in some wonky generic typing implicit Any situation I guess we're ok.

tlento (approved):
Looks good to me if this is what @QMalcolm needs!

The generics seem a little wonky here, but I guess it's the simplest annotation approach. Hopefully we won't uncover any weirdness, I always seem to find gaps in the ways generics get interpreted by mypy, and we end up getting away with things that shouldn't really be allowed.

dbt_semantic_interfaces/model_transformer.py
plypaul: This actually doesn't need the generic type. The `ProtocolHint` bit that's added later should clear it up.

PR #78, opened 2023-05-30
Add Protocol Hints

plypaul:
### Description

This PR adds `ProtocolHint` - a way to hint to the inspections in the IDE that a class is supposed to implement a `Protocol` without explicitly inheriting from a `Protocol`. The explicit inheritance can cause issues with Pydantic classes.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/semantic_manifest.py
+++ b/dbt_semantic_interfaces/implementations/semantic_manifest.py
@@ -2,14 +2,21 @@ from typing import List
 
 from importlib_metadata import version
 from pydantic import validator
+from typing_extensions import override
 
 from dbt_semantic_interfaces.implementations.base import HashableBaseModel
 from dbt_semantic_interfaces.implementations.metric import PydanticMetric
 from dbt_semantic_interfaces.implementations.semantic_model import PydanticSemanticModel
+from dbt_semantic_interfaces.protocols.protocol_hint import ProtocolHint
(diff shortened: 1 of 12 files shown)


dbt_semantic_interfaces/implementations/semantic_manifest.py
```diff
@@ -2,15 +2,22 @@
...
 
 
-class PydanticSemanticManifest(HashableBaseModel):
+class PydanticSemanticManifest(HashableBaseModel, ProtocolHint[SemanticManifest]):
     """Model holds all the information the SemanticLayer needs to render a query."""
 
+    @override
+    def _implements_protocol(self) -> SemanticManifest:  # noqa: D
+        return self
+
```
tlento, line 20: Oh I see, it's this approach. Yeah this is fine. I wish the toolchain just did this for us, but I can see why it's hard to do in the general case.

tlento (approved):
I wish the signatures weren't so clumsy but if it works, let's go with it for now. I like your idea of making this a class decorator, it's worth looking into that at some point.

plypaul:
> I like your idea of making this a class decorator, it's worth looking into that at some point.

@tlento I tried a few things to make that happen, but I couldn't get it to work. LMK if you have ideas.

tlento:
@plypaul bummer, I doubt I'd do anything different from what you tried.

PR #85, opened 2023-06-06
Add Validation for Semantic Model Defaults 

WilliamDee:
Resolves https://github.com/dbt-labs/metricflow/issues/565

### Description
Adds validation for checking the validity for `SemanticModel.defaults`. Currently, there's only `agg_time_dimension` which means it would validate whether the specified time dimension exists in the model.
 

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/semantic_models.py
+++ b/dbt_semantic_interfaces/validations/semantic_models.py
@@ -9,6 +9,7 @@ from dbt_semantic_interfaces.validations.validator_helpers import (
     FileContext,
     SemanticManifestValidationRule,
     SemanticModelContext,
+    SemanticModelValidationHelpers,
     ValidationError,
     ValidationIssue,
     validate_safely,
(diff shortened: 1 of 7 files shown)


dbt_semantic_interfaces/validations/semantic_models.py
```diff
@@ -178,3 +179,43 @@ def _validate_semantic_model(semantic_model: SemanticModel) -> List[ValidationIs
...
+    @staticmethod
+    @validate_safely(whats_being_done="checking validity of the semantic model's default agg_time_dimension")
+    def _validate_default_agg_time_dimension(semantic_model: SemanticModel) -> List[ValidationIssue]:
+        issues: List[ValidationIssue] = []
+
+        if semantic_model.defaults is None:
+            return []
+
+        default_agg_time_dimension = semantic_model.defaults.agg_time_dimension
+        assert default_agg_time_dimension, "should not be None"
```
QMalcolm, line 205: Nit: Because of python's wonkiness this assert could fail for other reasons than it being `None`. I think it's safer to specifically check `None`, i.e. `assert default_agg_time_dimension is not None`, simply because if we do run into the insanely weird scenario where `assert default_agg_time_dimension` fails for a reason other than `None`, we don't go the wrong direction when debugging.

QMalcolm (approved):
This looks amazing! I love that we made the change of where the `default_agg_time_dimension` is specified. And I extra love that we quick followed with this PR that has the validation change 😍

dbt_semantic_interfaces/validations/semantic_models.py
plypaul: Sorry, that was a mistake. It should also be optional in the Pydantic implementation.

QMalcolm: Hmmm. That's a great point. I personally think it should be optional (that then just means the user is required to specify an agg dimension for every measure on a given semantic model)

WilliamDee: cool, made it Optional!

PR #90, opened 2023-06-15
Use Metrics Instead of Measures as Inputs to Ratio Metrics

### LINKED ISSUE #67: [Spec Change] Update `RATIO` metric types to ingest metrics instead of measures

plypaul:
Resolves #67 

### Description

Implements the proposal in the linked issue.

### Checklist

### DIFF
--- a/.changes/unreleased/Breaking Changes-20230615-153301.yaml
+++ b/.changes/unreleased/Breaking Changes-20230615-153301.yaml
@@ -1,0 +1,6 @@
+kind: Breaking Changes
+body: Change ratio metrics to be configured with input metrics instead of measures.
+time: 2023-06-15T15:33:01.417121-07:00
+custom:
+  Author: plypaul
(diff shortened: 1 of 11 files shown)


QMalcolm (changes requested):
I think there is one change we need to make. In [schemas.py](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/c705896dda66460fd703a6da69ef04cfd70c009b/dbt_semantic_interfaces/parsing/schemas.py#L75) we list `measures` as a property for the metric type params. With this change, `measures` in the schemas should become `input_measures`. However, given that we set/override `input_measures` (and don't pull from it during that process), perhaps we should just remove it from the `schemas.py` because we don't want anyone to set it via their config.

plypaul:
Good point - updated.

QMalcolm (approved):
Send it!

PR #95, opened 2023-06-23
Add Configuration Object to Semantic Manifest

### LINKED ISSUE #94: [Feature] Add Configuration Object to Semantic Manifest

### Describe the feature

As requested by @marcodamore - it would be useful to have a configuration object in the semantic manifest to hold configuration options like:

```
{ 
  "project_config": {
      "dsi_version_number": 123,
      "timespine": {
          "location": "transform.metricflow_time_spine",
          "column_name": "date_day",
          "grain": "daily",
      }
   }
}
```

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

plypaul:
Resolves #94

### Description

Please see the linked issue.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/semantic_version.py
+++ b/dbt_semantic_interfaces/implementations/semantic_version.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+from typing import Optional
+
+from typing_extensions import override
+
+from dbt_semantic_interfaces.implementations.base import (
+    HashableBaseModel,
+    PydanticCustomInputParser,
+    PydanticParseableValueType,
+)
+from dbt_semantic_interfaces.protocols.semantic_version import SemanticVersion
+
+
+class PydanticSemanticVersion(PydanticCustomInputParser, HashableBaseModel):
+    """Pydantic implementation of SemanticVersion."""
+
+    @override
+    def _implements_protocol(self) -> SemanticVersion:
+        return self
+
+    major_version: str
+    minor_version: str
+    patch_version: Optional[str]
+
+    @classmethod
+    @override
+    def _from_yaml_value(cls, input: PydanticParseableValueType) -> PydanticSemanticVersion:
+        if isinstance(input, str):
+            return PydanticSemanticVersion.create_from_string(input)
+        else:
+            raise ValueError(
+                f"{cls.__name__} inputs from YAML files are expected to be of either type string or "
+                f"object (key/value pairs), but got type {type(input)} with value: {input}"
+            )
+
+    @staticmethod
+    def create_from_string(version_str: str) -> PydanticSemanticVersion:  # noqa: D
+        version_str_split = version_str.split(".")
+        if len(version_str_split) < 2:
+            raise ValueError(f"Expected version string to be of the form x.y or x.y.z, but got {version_str}")
+        return PydanticSemanticVersion(
+            major_version=version_str_split[0],
+            minor_version=version_str_split[1],
+            patch_version=".".join(version_str_split[2:]) if len(version_str_split) >= 3 else None,
+        )
+
+
+UNKNOWN_VERSION_SENTINEL = PydanticSemanticVersion(major_version="0", minor_version="0", patch_version="0")
--- a/dbt_semantic_interfaces/parsing/schemas.py
+++ b/dbt_semantic_interfaces/parsing/schemas.py
@@ -242,6 +242,33 @@ semantic_model_defaults_schema = {
 }
 
 
+time_spine_table_configuration_schema = {
+    "$id": "time_spine_table_configuration",
+    "type": "object",
+    "properties": {
+        "location": {"type": "string"},
+        "column_name": {"type": "string"},
+        "grain": {"enum": time_granularity_values},
+    },
+    "additionalProperties": False,
+    "required": ["location", "column_name", "grain"],
+}
+
+
+project_configuration_schema = {
+    "$id": "project_configuration",
+    "type": "object",
+    "properties": {
+        "time_spine_table_configurations": {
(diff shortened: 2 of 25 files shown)


tests/fixtures/semantic_manifest_yamls/simple_semantic_manifest/project_configuration.yaml
```diff
@@ -0,0 +1,7 @@
+---
+project_configuration:
+  dsi_package_version: 1.2.3
+  time_spine_table_configurations:
+    - location: example_schema.example_table
```
marcodamore, line 5: {{project_schema}}.mf_time_spine

dbt_semantic_interfaces/protocols/semantic_manifest.py
```diff
@@ -23,5 +24,10 @@ def metrics(self) -> Sequence[Metric]:  # noqa: D
     def interfaces_version(self) -> str:  # noqa: D
```
QMalcolm, line 24: Is this now duplicate information? Should we remove this property in favor of `dsi_package_version` in `project_configurations`?

QMalcolm (commented):
Looking good

dbt_semantic_interfaces/implementations/semantic_version.py
```diff
@@ -0,0 +1,47 @@
...
+    patch_version: Optional[str]
+
+    @classmethod
+    def _from_yaml_value(cls, input: PydanticParseableValueType) -> PydanticSemanticVersion:
+        """Parses a MetricInputMeasure from a string (name only) or object (struct spec) input.
+
+        For user input cases, the original YAML spec for a PydanticMetric included measure(s) specified as string names
+        or lists of string names. As such, configs pre-dating the addition of this model type will only provide the
+        base name for this object.
+        """
```
QMalcolm, line 33: This doc string need some updating

dbt_semantic_interfaces/implementations/semantic_version.py
```diff
@@ -0,0 +1,47 @@
...
+            return PydanticSemanticVersion(
+                major_version=input_split[0],
+                minor_version=input_split[1],
+                patch_version=input_split[2] if len(input_split) == 3 else None,
+            )
+        else:
+            raise ValueError(
+                f"{cls.__name__} inputs from YAML files are expected to be of either type string or "
+                f"object (key/value pairs), but got type {type(input)} with value: {input}"
+            )
```
QMalcolm, line 47: We probably want to update this to be a validator similar to https://github.com/dbt-labs/dbt-semantic-interfaces/blob/739fae883f991b92ad6db877223e5068a241947d/dbt_semantic_interfaces/implementations/semantic_manifest.py#L26

QMalcolm (changes requested):
More thoughts came to me after my initial review. We should never be expecting the user to to set the `dsi_package_version`. Thus we shouldn't be including it in the json schema spec. We should use a pydantic validator on the `PydanticSemanticVersion` implementation which parses it if a value is provided, otherwise it should populate the the current package version. The former change indicates to users that it isn't something they should directly set. The latter change allows for delineation between deserialization and new object instantiation.

dbt_semantic_interfaces/parsing/schemas.py
```diff
@@ -242,6 +242,34 @@
...
+    "additionalProperties": False,
+    "required": ["location", "column_name", "grain"],
+}
+
+
+project_configuration_schema = {
+    "$id": "project_configuration",
+    "type": "object",
+    "properties": {
+        "dsi_package_version": {"type": "string"},
```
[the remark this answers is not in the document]
plypaul: Can you clarify how that would work? e.g. we release DSI v0.2, but the customer wrote and committed configs based on DSI v0.1. If the server is upgraded and deserializes the files, wouldn't the version be incorrect?

dbt_semantic_interfaces/implementations/semantic_version.py
plypaul: Updated.

dbt_semantic_interfaces/parsing/schemas.py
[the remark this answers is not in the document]
QMalcolm: Deserialization and parsing logic is not the same. The JSON schema is only used for validating the yaml prior to parsing it to the object. Specifically it is invoked in the from of `<object_validator>.validate(...)` ([example](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/739fae883f991b92ad6db877223e5068a241947d/dbt_semantic_interfaces/parsing/dir_to_model.py#L297)). Deserialization is the act of calling `<object_type>.parse_raw(...yaml...)` deserialization by itself doesn't call json schema validation. Additionally in our json schema we explicitly disallow extra keys thus if someone tries to set a `dsi_package_version` we want things to fail. Because deserialization doesn't do json schema validation, this is not a problem. Thus by using using a pydantic `validate` setter we can handle both situtiations. When going through parsing, the `dsi_package_version` should be `None`, so we auto generate it. When deserializing, the value for `dsi_package_version` would be included in the serialization and thus will be used.

plypaul:
@QMalcolm - updated as per comments.

plypaul:
NVM, left out something.

dbt_semantic_interfaces/implementations/semantic_version.py
plypaul: A little unclear how the validator would work in the case with the custom parser. Mind elaborating?

QMalcolm (approved):
Looks great! Thank you so much for all the work!I know we had to go back and forth on a few things 🙂

PR #101, opened 2023-06-30
Fix minor issues with semantic model defaults validation

tlento:
The validation hook associated with SemanticModelDefaults would
register an error state if the optional agg_time_dimension property
was not set.

There was also a stray classmethod that really ought to be a staticmethod.
Forgive the "while I'm here" fixing but it doesn't seem worth a separate
commit.


QMalcolm (approved):
Ahh I see what was happening. The validation assumed that if defaults was specified, an agg time dim was required, when it's not. The change makes complete sense. Thank you for catching this and fixing it!

WilliamDee (approved):

PR #113, opened 2023-07-11
Switch from `checked_agg_time_dimension` to `checked_agg_time_dimension_for_measure`

QMalcolm:
Resolves #None

### Description

The `agg_time_dimension` of a measure is `Optional`, but we were only have treating it as such. @tlento's change in #101 was the first wave of this change. This proliferates the change fully thought DSI.

### Checklist


dbt_semantic_interfaces/implementations/semantic_model.py
```diff
@@ -167,3 +169,15 @@ def get_entity(self, entity_reference: LinkableElementReference) -> PydanticEnti
                 return entity
 
         raise ValueError(f"No entity with name ({entity_reference}) in semantic_model with name ({self.name})")
+
+    def checked_agg_time_dimension_for_measure(self, measure: Measure):  # noqa: D
```
plypaul, line 173: Instead of the measure object, using a measure reference seems more robust since this only applies to measures from this semantic model.

plypaul (approved):
LGTM aside from comment.

dbt_semantic_interfaces/implementations/semantic_model.py
QMalcolm: Agreed. Doing so adds the guarantee that the measure can be found on the semantic model. It's unfortunate that everywhere where we currently call `checked_agg_time_dimension` in `MetricFlow` and `dbt-semantic-interfaces` (and will be now calling `checked_agg_time_dimension_for_measure`) we already have the full measure object. So we'll be invoking another list iteration even though we don't have to. If that ever becomes an actual issue though, we'll know where to look.

Issue #120, opened 2023-07-20
Add `SemanticManifestValidator` rule for checking `WhereFilter.call_parameter_sets` of `Metric` filters


QMalcolm:
## Context
`Metric`s can have many `WhereFilter`s associated with them. Specifically:

- `Metric.filter`
- `Metric.type_params.measure.filter`
- `Metric.type_params.denominator.filter`
- `Metric.type_params.numerator.filter`
- `Metric.type_params.input_metrics[x].filter`
That's quite a few filters. Additionally the `where_sql_template` of a `WhereFilter` is a highly structured. Technically with #110 `call_parameter_sets` guarantees the structure of a `WhereFilter`s `where_sql_template`, but this only happens if `call_parameter_sets` is actually called. The best way to pseudo guarantee this happens is to add a [SemanticManifestValidationRule](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/707715b76d444483f3f1d2ef21b70561d983f659/dbt_semantic_interfaces/validations/validator_helpers.py#L389) to the [default rules](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/707715b76d444483f3f1d2ef21b70561d983f659/dbt_semantic_interfaces/validations/semantic_manifest_validator.py#L59) of the [SemanticManifestValidator](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/707715b76d444483f3f1d2ef21b70561d983f659/dbt_semantic_interfaces/validations/semantic_manifest_validator.py#L56). It's only a pseudo guarantee because from DSI's perspective, it's not guaranteed that a `SemanticManifest` has been run through the `SemanticManifestValidator`, however that is best practice and what people _should_ do.

## Acceptance Criteria
_**THERE EXISTS**_ a [SemanticManifestValidationRule](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/707715b76d444483f3f1d2ef21b70561d983f659/dbt_semantic_interfaces/validations/validator_helpers.py#L389)
_**THAT**_ checks `call_parameter_sets` of all filters of all metrics on a SemanticManifest
[...]

Issue #103, opened 2023-07-06
Add ValidationRule ensuring there are no duplicate primary entity parings


QMalcolm:
# Context / Problem

1. Multiple `SemanticModel` objects can have the same primary `Entity`
2. `Dimension` names are not required to unique across `SemanitcModel`s in the `SemanticManifest`

Together statements 1 and 2 mean that it is possible for `SemanticModel`s to have the same primary `Entity` and have `Dimension`s of the same name. This is problematic because if a dimension `<primary_entity>__<dimension_nam>` is specified in a `WhereFilter`, it can be ambiguous which dimension is actually being referenced.

# Solution

Disallow `SemanticModel`s with the same primary `Entity` to have duplicate `Dimension`s

# Accetance Criteria

A validation rule exists which disallows `SemanticModel`s with the same primary `Entity` to have duplicate `Dimension`s

PR #124, opened 2023-07-24
Add Primary Entity Field

### LINKED ISSUE #123: [Feature] Add Primary Entity Field For Local Dimensions

### Describe the feature

In a semantic model, there may be dimensions defined along with measures. These were called local dimensions, and they were accessed by the name of the element e.g. `bookings by is_instant`. However, this has resulted in a confusing interface as most other dimensions were accessed with an entity prefix. To make this interface more consistent, this issue proposes to add a `primary_entity` field to the semantic model that can be used to access those dimensions e.g. `bookings by booking__is_instant.`. If the semantic model contains entities as well, and there is an entity with the primary type, then those local dimensions can be accessed by that entity prefix, without setting the `primary_entity` field.

_No response_

### Who will this benefit?

_No response_

Yes.

### Anything else?

_No response_

plypaul:
Resolves #123 

### Description

Please see the linked issue for more details.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/semantic_model.py
+++ b/dbt_semantic_interfaces/implementations/semantic_model.py
@@ -171,12 +171,13 @@ class PydanticSemanticModel(HashableBaseModel, ModelWithMetadataParsing, Protoco
 
         raise ValueError(f"No entity with name ({entity_reference}) in semantic_model with name ({self.name})")
 
-    def checked_agg_time_dimension_for_measure(self, measure_reference: MeasureReference):  # noqa: D
+    def checked_agg_time_dimension_for_measure(  # noqa: D
+        self, measure_reference: MeasureReference
+    ) -> TimeDimensionReference:
         measure = self.get_measure(measure_reference=measure_reference)
-        if self.defaults is not None:
-            default_agg_time_dimesion = self.defaults.agg_time_dimension
+        default_agg_time_dimension = self.defaults.agg_time_dimension if self.defaults is not None else None
 
-        agg_time_dimension_name = measure.agg_time_dimension or default_agg_time_dimesion
+        agg_time_dimension_name = measure.agg_time_dimension or default_agg_time_dimension
         assert agg_time_dimension_name is not None, (
             f"Aggregation time dimension for measure {measure.name} is not set! This should either be set directly on "
             f"the measure specification in the model, or else defaulted to the primary time dimension in the data "
--- a/dbt_semantic_interfaces/implementations/semantic_model.py
+++ b/dbt_semantic_interfaces/implementations/semantic_model.py
@@ -19,6 +19,7 @@ from dbt_semantic_interfaces.protocols import (
     SemanticModelDefaults,
 )
 from dbt_semantic_interfaces.references import (
+    EntityReference,
     LinkableElementReference,
     MeasureReference,
     SemanticModelReference,
@@ -87,6 +88,7 @@ class PydanticSemanticModel(HashableBaseModel, ModelWithMetadataParsing, Protoco
     description: Optional[str]
     node_relation: NodeRelation
 
+    primary_entity: Optional[str]
     entities: Sequence[PydanticEntity] = []
     measures: Sequence[PydanticMeasure] = []
     dimensions: Sequence[PydanticDimension] = []
@@ -181,3 +183,7 @@ class PydanticSemanticModel(HashableBaseModel, ModelWithMetadataParsing, Protoco
             f"source containing the measure."
         )
         return TimeDimensionReference(element_name=agg_time_dimension_name)
+
+    @property
+    def primary_entity_reference(self) -> Optional[EntityReference]:  # noqa: D
+        return EntityReference(element_name=self.primary_entity) if self.primary_entity is not None else None
--- a/dbt_semantic_interfaces/protocols/semantic_model.py
+++ b/dbt_semantic_interfaces/protocols/semantic_model.py
@@ -8,6 +8,7 @@ from dbt_semantic_interfaces.protocols.entity import Entity
 from dbt_semantic_interfaces.protocols.measure import Measure
 from dbt_semantic_interfaces.protocols.metadata import Metadata
 from dbt_semantic_interfaces.references import (
+    EntityReference,
     LinkableElementReference,
     MeasureReference,
     SemanticModelReference,
@@ -71,6 +72,17 @@ class SemanticModel(Protocol):
     @property
     @abstractmethod
     def node_relation(self) -> NodeRelation:  # noqa: D
+        pass
+
+    @property
+    @abstractmethod
+    def primary_entity(self) -> Optional[str]:
+        """The primary entity for dimensions listed in this model.
+
+        This is for cases where there are dimensions in the model, but no entity with primary type. This allows those
+        dimensions to be accessed as dimensions need to be qualified by an entity for access. This may be None if there
+        are no dimensions in this model.
+        """
         pass
 
     @property
@@ -154,6 +166,12 @@ class SemanticModel(Protocol):
         Should raise an exception if a TimeDimensionReference cannot be built
         """
         ...
+
+    @property
+    @abstractmethod
+    def primary_entity_reference(self) -> Optional[EntityReference]:
+        """Reference object form of primary_entity."""
+        pass
 
 
 SemanticModelT = TypeVar("SemanticModelT", bound=SemanticModel)
--- a/dbt_semantic_interfaces/validations/primary_entity.py
+++ b/dbt_semantic_interfaces/validations/primary_entity.py
@@ -1,0 +1,103 @@
+import logging
+from typing import Generic, List, Sequence
+
+from dbt_semantic_interfaces.implementations.semantic_manifest import (
+    PydanticSemanticManifest,
+)
+from dbt_semantic_interfaces.protocols import SemanticManifestT, SemanticModel
+from dbt_semantic_interfaces.references import SemanticModelReference
+from dbt_semantic_interfaces.type_enums import EntityType
+from dbt_semantic_interfaces.validations.validator_helpers import (
+    FileContext,
+    SemanticManifestValidationRule,
+    SemanticModelContext,
+    ValidationError,
+    ValidationIssue,
+    validate_safely,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PrimaryEntityRule(SemanticManifestValidationRule[SemanticManifestT], Generic[SemanticManifestT]):
+    """Checks that the primary entity has been properly defined in a semantic model.
+
+    * If a semantic model contains dimensions, the primary entity must be available.
+    * The primary entity could be defined by the primary_entity field, or by one of the entities defined in a semantic
+      model.
+    * There should only be one primary entity in the model.
+    """
+
+    @staticmethod
+    def _model_requires_primary_entity(semantic_model: SemanticModel) -> bool:
+        return len(semantic_model.dimensions) > 0
+
+    @staticmethod
+    @validate_safely("Check that a semantic model has properly configured primary entities.")
+    def _check_model(semantic_model: SemanticModel) -> Sequence[ValidationIssue]:
+        context = SemanticModelContext(
+            file_context=FileContext.from_metadata(metadata=semantic_model.metadata),
+            semantic_model=SemanticModelReference(semantic_model_name=semantic_model.name),
+        )
+
+        # If there are entities defined in the model, check that there's only one primary entity.
+        entities_with_primary_type = tuple(
+            entity for entity in semantic_model.entities if entity.type is EntityType.PRIMARY
+        )
+
+        if len(entities_with_primary_type) > 0:
+            if len(entities_with_primary_type) > 1:
+                primary_entity_names = [primary_entity.name for primary_entity in entities_with_primary_type]
+                return (
+                    ValidationError(
+                        message=(
+                            f"Semantic models can have only one primary entity. The semantic model"
+                            f" `{semantic_model.name}` has {len(primary_entity_names)}: "
+                            f"{', '.join(primary_entity_names)}"
+                        ),
+                        context=context,
+                    ),
+                )
+
+            entity_with_primary_type = entities_with_primary_type[0]
+            # If there is a primary entity, the primary entity field should not be set.
+            if semantic_model.primary_entity_reference is not None:
+                return (
+                    ValidationError(
+                        message=(
+                            f"The semantic model `{semantic_model.name}` has an entity named "
+                            f"`{entity_with_primary_type.name}` with type primary but it also has the `primary_entity` "
+                            f"field set to `{semantic_model.primary_entity_reference.element_name}`. Both should not "
+                            f"be present in the model."
+                        ),
+                        context=context,
+                    ),
+                )
+
+        # Check that a primary entity has been set if required.
+        if (
+            PrimaryEntityRule._model_requires_primary_entity(semantic_model)
+            and semantic_model.primary_entity_reference is None
+            and len(entities_with_primary_type) == 0
+        ):
+            return (
+                ValidationError(
+                    message=(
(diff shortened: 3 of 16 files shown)


dbt_semantic_interfaces/implementations/semantic_model.py
```diff
@@ -169,15 +173,35 @@ def get_entity(self, entity_reference: LinkableElementReference) -> PydanticEnti
...
+            if entity.type is EntityType.PRIMARY:
+                return entity.reference
+            elif (
+                entity.type is EntityType.UNIQUE
+                or entity.type is EntityType.FOREIGN
+                or entity.type is EntityType.NATURAL
+            ):
+                pass
+            else:
+                assert_values_exhausted(entity.type)
```
QMalcolm, line 205: Is the `assert_values_exhausted` workflow actually necessary here? I think it's safe to assumethat the only entity type that this should ever return a reference for would be `EntityType.PRIMARY`, right? Like we don't think we're gonna add another entity type that should count as primary. If that is a safe assumption shouldn't we just do
```
if entity.type is EntityType.PRIMARY:
  return entity.reference
else:
  pass
```

dbt_semantic_interfaces/validations/primary_entity.py
```diff
@@ -0,0 +1,101 @@
...
+            entity for entity in semantic_model.entities if entity.type == EntityType.PRIMARY
+        )
+
+        if len(entities_with_primary_type) == 1:
+            return entities_with_primary_type[0]
+
+        raise RuntimeError(f"Did not find exactly one entity with entity type {EntityType.PRIMARY} in {semantic_model}")
+
+    @staticmethod
+    def _check_model(semantic_model: SemanticModel) -> Sequence[ValidationIssue]:
```
QMalcolm, line 52: This method should also have the `@validate_safely` decorator. By having it just on the base `validate_manifest` if one semantic model has an issue, any semantic models to be checked after it won't get checked because the exception handling happens outside of the for loop iterating over the models. By adding `@validate_safely` on `_check_model` it ensures that if an exception is raised for a semantic_model, it only stops this validation step for that semantic_model.

dbt_semantic_interfaces/validations/primary_entity.py
```diff
@@ -0,0 +1,101 @@
...
+                ValidationError(
+                    message=(
+                        f"The semantic model {semantic_model.name} contains dimensions, but it does not define a "
+                        f"primary entity."
+                    ),
+                    context=context,
+                ),
+            )
+
+        return []
```
QMalcolm, line 92: Is it not valid at all for a `SemanticModel` with dimensions to not have a primary entity, or is it something we simply want to discourage? If so should this instead be a `ValidationWarning`? Without a primary entity it just makes the dimensions in accessible to metrics based on different `SemanticModel`s right?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.1.0rc1"
+version = "0.1.0rc2"
```
QMalcolm, line 3: I think we should actually stop bumping the version as part of a feature PR. It's caused some discrepancies in the past. Releasing from a feature branch, releasing before the PR has been merged, etc. We should probably move to bumping the version as part of the release process, separate from any feature branch.

tests/fixtures/semantic_manifest_yamls/simple_semantic_manifest/semantic_models/bookings_source.yaml
```diff
@@ -94,6 +96,3 @@ semantic_model:
     - name: host
       type: foreign
       expr: host_id
-    - name: create_a_cycle_in_the_join_graph
-      type: primary
-      expr: guest_id
```
QMalcolm, line 99: This entity seems to exist specifically to create a cycle in the graph. Perhaps that is something we no longer care about in DSI, but I want to confirm that's the case.

dbt_semantic_interfaces/protocols/semantic_model.py
```diff
@@ -155,5 +156,14 @@ def checked_agg_time_dimension_for_measure(self, measure_reference: MeasureRefer
         """
         ...
 
+    @property
```
QMalcolm, line 159: Looks like we're missing `primary_entity: Optional[str]` on the protocol

QMalcolm (changes requested):
So this is technically a breaking change, and I'm wary that we should be doing that in the RC process, and this late in the RC process. Part of the problem is I'll have to see if I can get this into dbt-core, because in dbt-core we're only supposed to be doing bug fixes once we reach RC. I think we can get away with it though, and I believe you made that intentional? If so, nice 🙂  With `primary_entity` and `primary_entity_reference` are both `optional`, our tests in core checking that our implementations satisfy the protocols won't start failing. Additionally because `PrimaryEntityRule` wasn't added to the default `SemanticManifestValidator` rule set, we won't start running the validation rule in core, so calls to `primary_entity` and `primary_entity_reference` won't happen in core. However, with that said, does this help MetricFlow if we don't support it in core until 1.7? Would it be better to make these changes after we cut DSI 0.1.0 and start working on DSI 0.2.0. 

If we look at this repository as blind to MetricFlow and dbt-core (which I think we should generally strive for), then I'm not sure these changes make sense for 0.1.0 given where we are in the release process. If we're looking at this repository specifically in regards to dbt-core and MetricFlow then I'm also not sure that we should be moving forward with this now? Is there a benefit to having this in MF if we won't support it in core until 1.7? For clarity, I will see if we can make the necessary changes in core as part of the 1.6.0rc process.

dbt_semantic_interfaces/implementations/semantic_model.py
plypaul: This is the general pattern that we have been going with when switching on any enum to ensure type safety when an enum is changed. I'm sure there have been cases where we thought things would never change, but turned out to.

dbt_semantic_interfaces/validations/primary_entity.py
plypaul: It seems like an error to me. If the dimensions are not accessible - why have them defined at all? There could also be some poor UX here - a user makes a model, commits it (allowed because it's a warning), and then wonders why the dimensions aren't showing up.

dbt_semantic_interfaces/validations/primary_entity.py
plypaul: Updated.

tests/fixtures/semantic_manifest_yamls/simple_semantic_manifest/semantic_models/bookings_source.yaml
plypaul: That is a good case to check, but I ended up creating a separate model to test that case specifically. The issue was that we test for the output of dimensions frequently, and this name made it confusing to view the results. That's currently in MF - do we have a test here that would benefit from it?

dbt_semantic_interfaces/protocols/semantic_model.py
QMalcolm: If you look at this PR in isolation, no. However so far our approach has been that protocols have a number of attributes and some of those things should have a reference counter part that is exposed. We've actually discussed a few times whether the `*_reference` properties or if there should be a generic utility which can take in a DSI protocol object and expose the relevant `references`. Perhaps we decide later on to prioritize references in the protocol over the attributes that result in them, that references shouldn't be required properties of the protocol spec, or that both should continue to exist. But we need to make that decision, and we shouldn't being making a paradigm shift like that during the RC for a single attribute. DSI 0.1.0 should at least be consistent and thus for 0.1.0 exposes both.

dbt_semantic_interfaces/validations/primary_entity.py
QMalcolm: I think this is a misunderstanding on my part. I thought that the `Dimension`s would still be accessible by `Measure`s and `Metric`s that depend on the parent `SemanticModel`. Would it be valid for there to be a simple `Metric` based off a `Measure` in a semantic model that has a time `Dimension` that is the `Measure`s `agg_time_dimension` but no primary entity? An example

```
semantic_modesl:
  - name: user_created_event_log
    dimensions:
      - name: ds
        type: time
        expr: created_at
    measures:
      - name: users_created
        agg: COUNT
        expr: 1

metrics:
  - name: total users
     type: simple
     type_params:
         measure: users_created
```

plypaul:
> Regardless of the decision we eventually make, we should stay the current path for DSI 0.1.0.

Sounds good - updated.

dbt_semantic_interfaces/validations/primary_entity.py
plypaul: That would need a primary entity since the rule is: if there are dimensions, there needs to be a primary entity.

QMalcolm: When you say rule here I assume you mean how MetricFlow will be operating on things, not the validation rule itself. So MetricFlow would raise an exception at query time on above if attempting to query the metric `total_users`?

tests/fixtures/semantic_manifest_yamls/simple_semantic_manifest/semantic_models/bookings_source.yaml
QMalcolm: I think having the test in MF is a good place as it seems to make more sense as a SL implementation check than a protocol requirement.

dbt_semantic_interfaces/validations/primary_entity.py
```diff
@@ -0,0 +1,102 @@
...
+    SemanticModelContext,
+    ValidationError,
+    ValidationIssue,
+    validate_safely,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PrimaryEntityRule(SemanticManifestValidationRule[SemanticManifestT], Generic[SemanticManifestT]):
```
QMalcolm, line 22: We should get this added to the [DEFAULT_RULES](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/9cdf54593aa892fd969145bd53d7b2c891a6b62f/dbt_semantic_interfaces/validations/semantic_manifest_validator.py#L59) list if we want to ensure it's regularly enforced

dbt_semantic_interfaces/protocols/semantic_model.py
```diff
@@ -73,6 +74,17 @@ def description(self) -> Optional[str]:  # noqa: D
     def node_relation(self) -> NodeRelation:  # noqa: D
         pass
 
+    @property
+    @abstractmethod
+    def primary_entity(self) -> Optional[str]:  # noqa: D
```
QMalcolm, line 79: nit: given the docstring, I don't think we need the `# noqa: D` here

QMalcolm (commented):
We're still figuring out if this breaking change to the protocol spec should be included in DSI 0.1.0 and core 1.6.0. I believe we're leaning towards getting it in, should have an answer by tomorrow noon pacific time.

dbt_semantic_interfaces/validations/primary_entity.py
```diff
@@ -0,0 +1,102 @@
...
+        if (
+            PrimaryEntityRule._model_requires_primary_entity(semantic_model)
+            and semantic_model.primary_entity_reference is None
+            and not PrimaryEntityRule._model_has_entity_with_primary_type(semantic_model)
+        ):
+            return (
+                ValidationError(
+                    message=(
+                        f"The semantic model {semantic_model.name} contains dimensions, but it does not define a "
+                        f"primary entity."
```
tlento, line 87: Maybe update this to include more instructions? Something like this?

```suggestion
                        f"The semantic model {semantic_model.name} contains dimensions, but it does not define a "
                        f"primary entity. Either add an entity with type PRIMARY or set a value for the primary_entity key."
```

tlento (commented):
This seems reasonable to me.

Do we want to include the update to the WhereFilter parsing logic as part of this stack of PRs? Seems like we'd want that in place since it's a minimal risk change that allows us to avoid shipping a broken user experience, but if we do it needs to go out ASAP.

tlento:
Ok, so, we need:

1. Cleanup on this PR
2. 1 or 2 more validations (#103  and maybe #120 )
3. WhereFilter parsing logic update to take in the new style format

When do these all have to be in? Today, probably?

dbt_semantic_interfaces/validations/primary_entity.py
plypaul: Added.

dbt_semantic_interfaces/protocols/semantic_model.py
plypaul: Removed.

dbt_semantic_interfaces/validations/primary_entity.py
```diff
@@ -0,0 +1,102 @@
...
+            if entity.type == EntityType.PRIMARY:
+                return entity
+        entities_with_primary_type = tuple(
+            entity for entity in semantic_model.entities if entity.type == EntityType.PRIMARY
+        )
+
+        if len(entities_with_primary_type) == 1:
+            return entities_with_primary_type[0]
+
+        raise RuntimeError(f"Did not find exactly one entity with entity type {EntityType.PRIMARY} in {semantic_model}")
```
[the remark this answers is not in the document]
plypaul: My bad, iterated on these functions and this was from a previous implementation. Refactored a bit of this.

dbt_semantic_interfaces/validations/primary_entity.py
```diff
@@ -0,0 +1,102 @@
...
+        return any(entity.type == EntityType.PRIMARY for entity in semantic_model.entities)
+
+    @staticmethod
+    def _entity_with_primary_type_in_model(semantic_model: SemanticModel) -> Entity:
+        for entity in semantic_model.entities:
+            if entity.type == EntityType.PRIMARY:
+                return entity
+        entities_with_primary_type = tuple(
+            entity for entity in semantic_model.entities if entity.type == EntityType.PRIMARY
+        )
```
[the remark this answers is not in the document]
plypaul: As per previous reply, this has now been updated.

dbt_semantic_interfaces/validations/primary_entity.py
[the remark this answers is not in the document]
plypaul: Yeah, agree on the separate PR.

dbt_semantic_interfaces/validations/primary_entity.py
plypaul: Updated.

plypaul:
Updated with cleanup. Also rolled the `OnePrimaryEntityPerSemanticModelRule` into the `PrimaryEntity` rule, since it's kind of dependent.

dbt_semantic_interfaces/validations/primary_entity.py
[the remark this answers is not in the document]
QMalcolm: #127

dbt_semantic_interfaces/validations/primary_entity.py
```diff
@@ -0,0 +1,103 @@
...
+    @staticmethod
+    @validate_safely("Check that a semantic model has properly configured primary entities.")
+    def _check_model(semantic_model: SemanticModel) -> Sequence[ValidationIssue]:
+        context = SemanticModelContext(
+            file_context=FileContext.from_metadata(metadata=semantic_model.metadata),
+            semantic_model=SemanticModelReference(semantic_model_name=semantic_model.name),
+        )
+
+        # If there are entities defined in the model, check that there's only one primary entity.
+        entities_with_primary_time = tuple(
```
QMalcolm, line 44: nit: I'm confused why the word `time` is part of the variable name, is it meant to be `type`?

QMalcolm (approved):
I think there is one small variable name typo, but that's non blocking. This looks good to me! Thank you for doing all this work 🙏 I know we had a lot of back and forth on it.

dbt_semantic_interfaces/validations/primary_entity.py
plypaul: Yeah, should be. Updated.

PR #128, opened 2023-07-27
Add Primary Entity Prefix When Specifying Dimensions in the WhereFilter

### LINKED ISSUE #123: [Feature] Add Primary Entity Field For Local Dimensions

plypaul:
Resolves #123

### Description

With the invariant that all dimensions are associated with a primary entity, require the specification of the entity when using dimensions in the where filter.

e.g.

```
dimension('capacity_latest') > 10

->

dimension('listing__capacity_latest') > 10
```

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/naming/dundered.py
+++ b/dbt_semantic_interfaces/naming/dundered.py
@@ -1,0 +1,138 @@
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import Optional, Tuple
+
+from dbt_semantic_interfaces.naming.keywords import DUNDER
+from dbt_semantic_interfaces.references import EntityReference
+from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class StructuredDunderedName:
+    """Group by items (e.g. dimensions / entities) in a query that are named using a double underscore as a seperator.
+
+    e.g. listing__ds__week ->
+    entity_links: ["listing"]
+    element_name: "ds"
+    granularity: TimeGranularity.WEEK
+
+    The time granularity is part of legacy query syntax and there are plans to migrate away from this format.
+    """
+
+    entity_links: Tuple[EntityReference, ...]
+    element_name: str
+    time_granularity: Optional[TimeGranularity] = None
+
+    @staticmethod
+    def parse_name(name: str) -> StructuredDunderedName:
+        """Construct from a string like 'listing__ds__month'."""
+        name_parts = name.split(DUNDER)
+
+        # No dunder, e.g. "ds"
+        if len(name_parts) == 1:
+            return StructuredDunderedName((), name_parts[0])
+
+        associated_granularity = None
+        granularity: TimeGranularity
+        for granularity in TimeGranularity:
+            if name_parts[-1] == granularity.value:
+                associated_granularity = granularity
+
+        # Has a time granularity
+        if associated_granularity:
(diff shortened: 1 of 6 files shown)


dbt_semantic_interfaces/naming/dundered.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import Optional, Tuple
+
+from dbt_semantic_interfaces.references import EntityReference
+from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
+
+DUNDER = "__"
```
QMalcolm, line 10: Instead of defining this again here, shouldn't we import the `DUNDER` defined in `keywords.py`?

dbt_semantic_interfaces/parsing/where_filter_parser.py
```diff
@@ -38,10 +50,19 @@ def parse_call_parameter_sets(where_sql_template: str) -> FilterCallParameterSet
...
+                )
+
             dimension_call_parameter_sets.append(
                 DimensionCallParameterSet(
-                    dimension_reference=DimensionReference(element_name=dimension_name),
-                    entity_path=tuple(EntityReference(element_name=arg) for arg in entity_path),
+                    dimension_reference=DimensionReference(element_name=group_by_item_name.element_name),
+                    entity_path=(
+                        tuple(EntityReference(element_name=arg) for arg in entity_path)
+                        + group_by_item_name.entity_links
```
QMalcolm, line 64: So the primary entity will be the last one in the entity_path tuple. Cool Cool 👍

QMalcolm (approved):
One nit, but otherwise looks wonderful!

dbt_semantic_interfaces/naming/dundered.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@dataclass(frozen=True)
+class DunderedNameFormatter:
+    """Group by items (e.g. dimensions / entities) in a query that are named using a double underscore as a seperator.
+
+    e.g. listing__ds__week ->
+    entity_links: ["listing"]
+    element_name: "ds"
+    granularity: TimeGranularity.WEEK
```
tlento, line 22: Can we indicate that including time granularity is a legacy anti-pattern and no implementer should be relying on it? It has caused so much trouble, and we are trying to eliminate it somehow or other.

tlento (commented):
YAY!

I don't like time granularity dunder joins but it's better to have them here at this time. Let's just annotate it as undesirable in the docstring.

dbt_semantic_interfaces/naming/dundered.py
plypaul: Refactor error - updated.

dbt_semantic_interfaces/naming/dundered.py
plypaul: Added.

PR #131, opened 2023-07-28
Add `WhereFiltersAreParseable` Validation Rule

### LINKED ISSUE #120: Add `SemanticManifestValidator` rule for checking `WhereFilter.call_parameter_sets` of `Metric` filters

QMalcolm:
Resolves #120

### Description

The structure of the `where_sql_template` of `WhereFilter`s is highly structured. However nothing was actually enforcing that structure. This PR adds a validation which calls `call_parameter_sets` on `WhereFilter`s which tries to parse the `where_sql_template`. Validations aren't required to be run, although recommended. So this is more of soft protocol guarantee.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/metrics.py
+++ b/dbt_semantic_interfaces/validations/metrics.py
@@ -17,6 +17,7 @@ from dbt_semantic_interfaces.validations.validator_helpers import (
     SemanticManifestValidationRule,
     ValidationError,
     ValidationIssue,
+    generate_exception_issue,
     validate_safely,
 )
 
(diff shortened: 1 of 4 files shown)


dbt_semantic_interfaces/validations/metrics.py
```diff
@@ -153,3 +154,113 @@ def validate_manifest(semantic_manifest: SemanticManifestT) -> Sequence[Validati
...
+
+        if metric.type_params:
+            measure = metric.type_params.measure
+            if measure is not None and measure.filter is not None:
+                try:
+                    measure.filter.call_parameter_sets
+                except Exception as e:
+                    issues.append(
+                        generate_exception_issue(
+                            what_was_being_done=f"trying to parse the measure filter on metric `{metric.name}`",
```
tlento, line 197: Shall we include the measure name here?
```suggestion
                            what_was_being_done=f"trying to parse the filter for measure `{measure.name}` on metric `{metric.name}`",
```

dbt_semantic_interfaces/validations/metrics.py
```diff
@@ -153,3 +154,113 @@ def validate_manifest(semantic_manifest: SemanticManifestT) -> Sequence[Validati
...
+                    )
+
+            for derived_metric in metric.type_params.metrics or []:
+                if derived_metric.filter is not None:
+                    try:
+                        derived_metric.filter.call_parameter_sets
+                    except Exception as e:
+                        issues.append(
+                            generate_exception_issue(
+                                what_was_being_done=f"trying to parse derived metric filter on metric `{metric.name}`",
```
tlento, line 248: This is confusing for the user, because it's the input metric, not the derived metric:

```suggestion
                                what_was_being_done=f"trying to parse filter for input metric `{derived_metric.name}` on metric `{metric.name}`",
```

tlento (approved):
LGTM, thanks! I have a couple of nits inline on messaging. Note the pytest.raises() match parameter takes a regex so the updates shouldn't be too painful.

dbt_semantic_interfaces/validations/metrics.py
```diff
@@ -153,3 +154,113 @@ def validate_manifest(semantic_manifest: SemanticManifestT) -> Sequence[Validati
...
+    def _validate_metric(metric: Metric) -> Sequence[ValidationIssue]:  # noqa: D
+        issues: List[ValidationIssue] = []
+        context = MetricContext(
+            file_context=FileContext.from_metadata(metadata=metric.metadata),
+            metric=MetricModelReference(metric_name=metric.name),
+        )
+
+        if metric.filter is not None:
+            try:
+                metric.filter.call_parameter_sets
```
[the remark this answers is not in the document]
QMalcolm: Agreed, it's not my preferred path. As it is `call_parameter_sets` is what guarantees the structure of the `where_sql_template`, and thus this is really the only way to do it 😞

dbt_semantic_interfaces/validations/metrics.py
QMalcolm: I don't actually think thats necessary. `measure` is directly a `type_params` attribute which is where the filter lives. I worry adding the measure name would actually cause confusion wherein people would look at a measure on a semantic model instead of addressing the measure of the metrics `type_params`

QMalcolm: Never mind, you're right, I agree 😂

PR #148, opened 2023-09-14
Add New SavedQuery Protocol

### LINKED ISSUE #144: Add new `SavedQuery` protocol
In https://github.com/dbt-labs/metricflow/issues/765 new functionality for saved queries will be introduced in MetricFlow. To support this we need to extend the [SemanticManifest](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/ef8a893cc6e65f695db46356b8471a039246456b/dbt_semantic_interfaces/protocols/semantic_manifest.py#L9) protocol to have a new top level attribtue `saved_queries` which will be a sequence of `SavedQuery` protocols.

The `SavedQuery` protocol will need to have the following attributes

- `name` a string
- `description` an optional string
- `query` a sub protocol `Query`

The `Query` protocol will need to have the following attributes

- `metrics` an (optional?) list of metrics
- `group_bys` an (optional?) list of ?
- `where` an (optional?) ?
- `limit` an (optional?) integer
- `order` an (optional?) ?

plypaul:
Resolves #144 

### Description

Please see the associated issue for details.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/saved_query.py
+++ b/dbt_semantic_interfaces/implementations/saved_query.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+from typing import List, Optional
+
+from typing_extensions import override
+
+from dbt_semantic_interfaces.implementations.base import (
+    HashableBaseModel,
+    ModelWithMetadataParsing,
+)
+from dbt_semantic_interfaces.implementations.filters.where_filter import (
+    PydanticWhereFilter,
+)
+from dbt_semantic_interfaces.implementations.metadata import PydanticMetadata
+from dbt_semantic_interfaces.protocols import ProtocolHint
+from dbt_semantic_interfaces.protocols.saved_query import SavedQuery
+
+
+class PydanticSavedQuery(HashableBaseModel, ModelWithMetadataParsing, ProtocolHint[SavedQuery]):
+    """Pydantic implementation of SavedQuery."""
+
+    @override
+    def _implements_protocol(self) -> SavedQuery:
+        return self
+
+    name: str
+    metrics: List[str]
+    group_bys: List[str] = []
+    where: List[PydanticWhereFilter] = []
+
+    description: Optional[str] = None
+    metadata: Optional[PydanticMetadata] = None
--- a/dbt_semantic_interfaces/parsing/schemas.py
+++ b/dbt_semantic_interfaces/parsing/schemas.py
@@ -269,6 +269,29 @@ project_configuration_schema = {
 }
 
 
+saved_query_schema = {
+    "$id": "saved_query_schema",
+    "type": "object",
+    "properties": {
+        "name": {"type": "string"},
+        "description": {"type": "string"},
+        "metrics": {
+            "type": "array",
+            "items": {"type": "string"},
+        },
+        "group_bys": {
+            "type": "array",
+            "items": {"type": "string"},
+        },
+        "where": {
+            "type": "array",
+            "items": {"type": "string"},
+        },
+    },
+    "required": ["name", "metrics"],
+    "additionalProperties": False,
+}
+
 semantic_model_schema = {
     "$id": "semantic_model_schema",
     "type": "object",
@@ -297,6 +320,7 @@ schema_store = {
     metric_schema["$id"]: metric_schema,
     semantic_model_schema["$id"]: semantic_model_schema,
     project_configuration_schema["$id"]: project_configuration_schema,
+    saved_query_schema["$id"]: saved_query_schema,
     # Sub-object schemas
     metric_input_measure_schema["$id"]: metric_input_measure_schema,
     metric_type_params_schema["$id"]: metric_type_params_schema,
@@ -318,3 +342,4 @@ resolver = RefResolver.from_schema(schema=metric_schema, store=schema_store)
 semantic_model_validator = SchemaValidator(semantic_model_schema, resolver=resolver)
 metric_validator = SchemaValidator(metric_schema, resolver=resolver)
 project_configuration_validator = SchemaValidator(project_configuration_schema, resolver=resolver)
+saved_query_validator = SchemaValidator(saved_query_schema, resolver=resolver)
--- a/dbt_semantic_interfaces/protocols/saved_query.py
+++ b/dbt_semantic_interfaces/protocols/saved_query.py
@@ -1,0 +1,39 @@
+from abc import abstractmethod
+from typing import Optional, Protocol, Sequence
+
+from dbt_semantic_interfaces.protocols.metadata import Metadata
+from dbt_semantic_interfaces.protocols.where_filter import WhereFilter
+
+
+class SavedQuery(Protocol):
+    """Represents a query that the user wants to run repeatedly."""
+
+    @property
+    @abstractmethod
+    def metadata(self) -> Optional[Metadata]:  # noqa: D
+        pass
+
+    @property
+    @abstractmethod
+    def name(self) -> str:  # noqa: D
+        pass
+
+    @property
+    @abstractmethod
+    def description(self) -> Optional[str]:  # noqa: D
+        pass
+
+    @property
+    @abstractmethod
+    def metrics(self) -> Sequence[str]:  # noqa: D
+        pass
+
+    @property
+    @abstractmethod
+    def group_bys(self) -> Sequence[str]:  # noqa: D
+        pass
+
+    @property
+    @abstractmethod
+    def where(self) -> Sequence[WhereFilter]:  # noqa: D
+        pass
--- a/dbt_semantic_interfaces/protocols/semantic_manifest.py
+++ b/dbt_semantic_interfaces/protocols/semantic_manifest.py
@@ -3,6 +3,7 @@ from typing import Protocol, Sequence, TypeVar
 
 from dbt_semantic_interfaces.protocols.metric import Metric
 from dbt_semantic_interfaces.protocols.project_configuration import ProjectConfiguration
+from dbt_semantic_interfaces.protocols.saved_query import SavedQuery
 from dbt_semantic_interfaces.protocols.semantic_model import SemanticModel
 
 
@@ -22,6 +23,10 @@ class SemanticManifest(Protocol):
     @property
     @abstractmethod
     def project_configuration(self) -> ProjectConfiguration:  # noqa: D
+        pass
+
+    @property
+    def saved_queries(self) -> Sequence[SavedQuery]:  # noqa: D
         pass
 
 
--- a/dbt_semantic_interfaces/validations/saved_query.py
+++ b/dbt_semantic_interfaces/validations/saved_query.py
@@ -1,0 +1,146 @@
+import logging
+import traceback
+from typing import Generic, List, Sequence, Set
+
+from dbt_semantic_interfaces.call_parameter_sets import FilterCallParameterSets
+from dbt_semantic_interfaces.naming.keywords import METRIC_TIME_ELEMENT_NAME
+from dbt_semantic_interfaces.parsing.where_filter_parser import WhereFilterParser
+from dbt_semantic_interfaces.protocols import SemanticManifestT
+from dbt_semantic_interfaces.protocols.saved_query import SavedQuery
+from dbt_semantic_interfaces.validations.validator_helpers import (
+    FileContext,
+    SavedQueryContext,
+    SavedQueryElementType,
+    SemanticManifestValidationRule,
+    ValidationError,
+    ValidationIssue,
+    generate_exception_issue,
+    validate_safely,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class SavedQueryRule(SemanticManifestValidationRule[SemanticManifestT], Generic[SemanticManifestT]):
+    """Validates fields in a saved query.
+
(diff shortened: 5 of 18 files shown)


dbt_semantic_interfaces/protocols/saved_query.py
```diff
@@ -0,0 +1,39 @@
+from abc import abstractmethod
+from typing import Optional, Protocol, Sequence
+
+from dbt_semantic_interfaces.protocols.metadata import Metadata
+from dbt_semantic_interfaces.protocols.where_filter import WhereFilter
+
+
+class SavedQuery(Protocol):
```
QMalcolm, line 8: We've got separate work that is to add an optional `label` to many objects (metrics, semantic_models, dimensions, measures). Are we planning to add them to saved queries as well? If so we could do so now. It'll make it easier on the core side of things because we won't have to implement saved queries, and then come back and add a new attribute to them a week later.

dbt_semantic_interfaces/protocols/saved_query.py
```diff
@@ -0,0 +1,39 @@
...
+        pass
+
+    @property
+    @abstractmethod
+    def metric_names(self) -> Sequence[str]:  # noqa: D
+        pass
+
+    @property
+    @abstractmethod
+    def group_by_item_names(self) -> Sequence[str]:  # noqa: D
```
QMalcolm, line 33: I assume the idea is that `group_by_item_names` aren't required, but should be represented as an empty set when there are none? Was there a conversation somewhere about whether not we should mark them as optional (`Optional[Sequence[str]]`)? I'd just like to understand the logic so I can keep it in mind when reviewing other parts of the protocol and ensuring consistency.

dbt_semantic_interfaces/protocols/semantic_manifest.py
```diff
@@ -24,5 +25,13 @@ def metrics(self) -> Sequence[Metric]:  # noqa: D
     def project_configuration(self) -> ProjectConfiguration:  # noqa: D
         pass
 
+    @property
+    def saved_queries(self) -> Sequence[SavedQuery]:  # noqa: D
+        """Saved queries defined in the manifest.
+
+        Returns a default value for backward compatibility until the next breaking release.
+        """
+        return ()
```
QMalcolm, line 34: Hmmm 🤔 do we need this return? Nothing should be directly inheriting the protocol, and thus won't inherit this implementation. It's up to the implementation of protocols to define the default behavior to satisfy the typing. We should probably stick to the pattern of just having `pass` in the protocol definitions

dbt_semantic_interfaces/validations/saved_query.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class SavedQueryRule(SemanticManifestValidationRule[SemanticManifestT], Generic[SemanticManifestT]):
+    """Validates fields in a saved query.
+
+    As the semantic model graph is not traversed completely in DSI, the validations for saved queries can't be complete.
+    Consequently, the current plan is that we add a separate validation using MetricFlow in CI.
+
+    * Check if metric names exist in the manifest.
+    * Check that the where filter is valid using the same logic as WhereFiltersAreParsable
```
QMalcolm, line 28: Do we also need to check `group_by_item_names` to check that they reference existing things? Additionally, what are valid `group_by_item_names`, is it just categorical dimensions?

QMalcolm (changes requested):
This looks fantastic! Some small questions that we should figure out before approving and merging. Just want to have things really nailed down before starting the core work.

dbt_semantic_interfaces/protocols/saved_query.py
tlento: As a general rule, collections should not be typed as optional unless there's a meaningful difference between "no collection" and "empty collection".

So we should not make these optional return types, but on implementation we can default them to an empty Sequence type so users are not required to pass in a value.

dbt_semantic_interfaces/implementations/saved_query.py
```diff
@@ -0,0 +1,40 @@
...
+class PydanticSavedQuery(HashableBaseModel, ModelWithMetadataParsing, ProtocolHint[SavedQuery]):
+    """Pydantic implementation of SavedQuery."""
+
+    @override
+    def _implements_protocol(self) -> SavedQuery:
+        return self
+
+    name: str
+    metrics: List[str]
+    group_bys: List[str]
```
tlento, line 28: @QMalcolm does it matter that this is not part of the protocol? It shouldn't, but I wanted to make sure this won't cause havoc with the core parsing setup.

dbt_semantic_interfaces/protocols/semantic_manifest.py
tlento: +1 @QMalcolm  - returning a value from the protocol affords zero protection against breakage, since nothing has to inherit from it, and given that our position is you shouldn't inherit from protocols in the general case I agree we should leave the implementations out of here.

dbt_semantic_interfaces/protocols/saved_query.py
```diff
@@ -0,0 +1,39 @@
...
+        pass
```
tlento, line 34: Separate question, why do we need this named in this way? Is this for internal developer ergonomics?

It's a little odd to me to take in group_bys as a list of strings and just return it from the more verbose protocol named field.

Also, how does this work with the consolidation we're trying to do around how people reference these things? Like what happens with `Dimension(.....)` calls? Do we want to make this conform to the interface we ultimately want instead of sticking with bare string references everywhere?

dbt_semantic_interfaces/protocols/saved_query.py
QMalcolm: Perfect, that works for me 🙂

dbt_semantic_interfaces/implementations/saved_query.py
QMalcolm: Good catch. It would make things weird. To stick to the protocol we'd have to serialize `group_by_item_names`, but then MetricFlow uses these pydantic objects for deserialization, and since their is no setter for `group_by_item_names` on this object, they'd get dropped 😬 In a separate comment you brought up a question about why do we have both `group_bys` and `group_by_item_names`, and I think that's actually the better question. I think we just want `group_bys`? I'm not seeing what `group_by_item_names` provides.

dbt_semantic_interfaces/protocols/semantic_manifest.py
plypaul: Yeah, if nothing is inheriting from it, I'll remove.

dbt_semantic_interfaces/protocols/saved_query.py
plypaul: Yeah, it was for developer ergonomics. The direction that @QMalcolm mentioned that we wanted to head to was decouple the protocol from the YAML. But seems like that's not the case. @QMalcolm mind elaborating?

dbt_semantic_interfaces/protocols/saved_query.py
plypaul: Not sure about whether it's supposed to be added, but seems fine to.

dbt_semantic_interfaces/protocols/saved_query.py
QMalcolm: We do want to head the direction of decoupling the protocol from the YAML. Unfortunately `dbt-semantic-interfaces` does not currently have an architecture to support that. Specifically the json schema definition has to map to the pydantic implementation, and the pydantic implementation must satisfy the protocol. This architecture also leads us to marking things as optional in the protocol that are actually required, because the implementation has to mark them as optional (but we always ensure they're filled in with transformation rules).

Interestingly `dbt-core` does have an architecture that allows decoupling the yaml from the protocol. In dbt-core there is a concept of unparsed nodes and parsed nodes. YAML is loaded raw into the unparsed nodes. Then there exists mapping logic (similar to what DSI's transformation rules do) that create the "full" parsed node which satisfies the protocol. This is the direction I think we should move to in DSI, and I think it's out of scope for this PR.

The specific problem with this implementation is that it doesn't allow for setting `group_by_item_names` in deserialization. This actually ties us to the yaml because we'd actually have to output `group_bys` in the serialization, however that isn't part of the protocol so it shouldn't be part of the serialization. Could you dive into deeper what ergonomics are provided by having both `group_bys` and `group_by_item_names` as they both return the same thing currently?

plypaul: Got it. Regarding the naming - it was just for consistency with how things were named in MF. Given the situation, I'm fine removing `group_by_item_names`.

dbt_semantic_interfaces/validations/saved_query.py
plypaul: @QMalcolm Yeah, that would be good to check, but I didn't see similar checks being done for filters. Is that right?

QMalcolm: Oh right group_bys take the form of `Dimension(...)`. Then we need a validation similar to the `WhereFilter` validations that checks that they're parsable by the jinja templater, because as with `WhereFilters` the jinja template parser is the only thing can enforce the protocol structure of these structured strings.

dbt_semantic_interfaces/protocols/saved_query.py
QMalcolm: I think we'll just quickly add it in a PR where we add label to everything else. So don't worry about it here.

dbt_semantic_interfaces/validations/saved_query.py
```diff
@@ -29,14 +31,30 @@ class SavedQueryRule(SemanticManifestValidationRule[SemanticManifestT], Generic[
...
 
     @staticmethod
-    def _model_requires_primary_entity(semantic_model: SemanticModel) -> bool:
-        return len(semantic_model.dimensions) > 0
+    @validate_safely("Validate the group-by field in a saved query.")
+    def _check_group_bys(valid_group_by_element_names: Set[str], saved_query: SavedQuery) -> Sequence[ValidationIssue]:
+        issues: List[ValidationIssue] = []
+
+        for group_by_item_name in saved_query.group_bys:
+            structured_name = DunderedNameFormatter.parse_name(group_by_item_name)
```
QMalcolm, line 39: We already talked about it in a call, but documenting it here. This assumes that the `group_bys` will be strings of the form `'<primary_enitity>__<dimension>'`. Although eventually we want to support that, the strings right now will be of the form `'Dimension("<primary_entity>__<dimension>")'`. We should probably have a `GroupByParser`, similar to the `WhereFilterParser` for parsing group_bys, and use that for validating that the group_bys have valid structuring.

QMalcolm (changes requested):
We're looking really good, we just need to do a bit more on handing of group_bys I believe.

dbt_semantic_interfaces/validations/saved_query.py
plypaul: Re-used the `WhereFilterParser` for now - there's some work going on right now to support `Dimension(...).grain(...)`, so I'll wait until that's in place to switch out the implementation.

dbt_semantic_interfaces/protocols/semantic_manifest.py
QMalcolm: This still should changed to `pass`

dbt_semantic_interfaces/validations/saved_query.py
QMalcolm: Changes to this look good!

dbt_semantic_interfaces/implementations/saved_query.py
```diff
@@ -0,0 +1,32 @@
...
```
QMalcolm, line 28: We probably should default `group_bys` to an empty list as I don't believe they're required.

dbt_semantic_interfaces/parsing/schemas.py
```diff
@@ -269,6 +269,28 @@
...
+        "group_bys": {
+            "type": "array",
+            "items": {"type": "string"},
+        },
+        "where": {
+            "type": "array",
+            "items": {"type": "string"},
+        },
+    },
+    "additionalProperties": False,
```
QMalcolm, line 291: Nit: I don't think we have to, but we should probably mark the properties `name` and `metrics` to be required. I think parsing will error out either way if its not specified, as the creation of the object will fail. But specifying them in the jsonschema spec will give better errors I believe.

QMalcolm (changes requested):
Some small changes, left, but the overall implementation looks great!

dbt_semantic_interfaces/parsing/schemas.py
plypaul: Updated.

dbt_semantic_interfaces/implementations/saved_query.py
plypaul: Updated.

dbt_semantic_interfaces/protocols/semantic_manifest.py
plypaul: Missed earlier, but now updated.

QMalcolm (approved):
Looks good to me! Sorry that there was so much back and forth on this one 🙃

PR #150, opened 2023-09-14
Update valid name regex to disallow dunders

### LINKED ISSUE #149: [Bug] UniqueAndValidNameRule allows DUNDER inside of element names

### Current Behavior

Our validation rules allow users to specify things like entity names with the DUNDER (`__` which is a double underscore) inside of them.

Our query parser and all of our linkable spec handling (and, for the moment, granularity labeling) assumes that there are no DUNDER sequences in user-land names.

This means the configs will pass validation, but then all kinds of weird stuff can happen at query time with inscrutable failures.

### Expected Behavior

The validation should just block these names with a helpful error message.

### Steps To Reproduce

1. Create an entity with type: primary and a name containing a double-underscore somewhere in the middle
2. Try to run a query with a properly formed filter requesting any dimension from the semantic model with your silently broken entity (using the `{{ Dimension('....') }}` syntax)
3. Observe the not at all hilarious failure mode where we parse that dimension name and blow up saying it's malformed.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- dbt-semantic-interfaces:
```

### Additional Context

_No response_

QMalcolm:
Resolves #149 

### Description

We use dunders `__` to separate named objects in a single string reference. I.e. `orders__is_large_order` where `order` is the primary entity and `is_large_order` is the dimension name. If people use dunders in their names, then this logic breaks. This PR updates the regex for name validity checking to disallow dunders.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/unique_valid_name.py
+++ b/dbt_semantic_interfaces/validations/unique_valid_name.py
@@ -73,9 +73,10 @@ class UniqueAndValidNameRule(SemanticManifestValidationRule[SemanticManifestT], 
             issues.append(
                 ValidationError(
                     context=context,
-                    message=f"Invalid name `{name}` - names should only consist of lower case letters, numbers, "
-                    f"and underscores. In addition, names should start with a lower case letter, and should not end "
-                    f"with an underscore, and they must be at least 2 characters long.",
+                    message=f"Invalid name `{name}` - names may only contain lower case letters, numbers, "
+                    f"and underscores. Additionally, names must start with a lower case letter, cannot end "
+                    f"with an underscore, cannot contain dunders (double underscores, or __), and must be "
+                    f"at least 2 characters long.",
                 )
             )
         if name.upper() in TimeGranularity.list_names():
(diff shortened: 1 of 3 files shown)


dbt_semantic_interfaces/validations/unique_valid_name.py
```diff
@@ -69,9 +73,9 @@ def check_valid_name(name: str, context: Optional[ValidationContext] = None) ->
             issues.append(
                 ValidationError(
                     context=context,
-                    message=f"Invalid name `{name}` - names should only consist of lower case letters, numbers, "
-                    f"and underscores. In addition, names should start with a lower case letter, and should not end "
-                    f"with an underscore, and they must be at least 2 characters long.",
+                    message=f"Invalid name `{name}` - names may only contain lower case letters, numbers, "
+                    f"and underscores. Additionally, names must start with a lower case letter, cannot end "
+                    f"with an underscore, cannot contain dunders (__), and must be at least 2 characters long.",
```
tlento, line 78: nit: DUNDER and __ might not render in an obvious way.
```suggestion
                    f"with an underscore, cannot contain dunders (double underscores, or __), and must be at least 2 characters long.",
```

tlento (commented):
Do we want this new rule to apply to semantic model names? I think we do. But that may cause some breakage for things which work today.

tlento (approved):

dbt_semantic_interfaces/validations/unique_valid_name.py
QMalcolm: Added the suggested change via a fixup to "Improve validation error message for invalid names" commit, and force pushed.

PR #152, opened 2023-09-18
Support for `Dimension.grain(...)` in where/filter

DevonFulcher:
Resolves #SL-631 (internal to dbt Labs)

### Description

The object of this PR is to allow for syntax like this in the where parameter & filter field `{{ Dimension('metric_time').grain('WEEK') }} > 2023-09-18`. To accomplish this, I refactored `WhereFilterParser` to use the `QueryInterface*` classes such as `QueryInterfaceDimension`. This has been done in MF already for [WhereSpecFactory](https://github.com/dbt-labs/metricflow/blob/main/metricflow/specs/where_filter_transform.py).

Now that MF, MFS, and DSI all implement the same `QueryInterface*` classes, there will be a lessened likelihood of drift between these codebases. I plan on removing the `QueryInterface*` classes from MF and having that codebase reference the protocols in the codebase instead. 

### Checklist


dbt_semantic_interfaces/parsing/where_filter/query_interface.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import annotations
+
+from typing import Protocol, Sequence
+
+
```
QMalcolm, line 5: Given that this file solely contains protocols, what are people's thoughts on moving it the the `protocols` directory? cc: @plypaul & @tlento

dbt_semantic_interfaces/parsing/where_filter/query_interface.py
```diff
@@ -0,0 +1,65 @@
...
+class QueryInterfaceDimension(Protocol):
+    """Represents the interface for Dimension in the query interface."""
+
+    def grain(self, _grain: str) -> QueryInterfaceDimension:
+        """The time granularity."""
+        raise NotImplementedError
```
QMalcolm, line 11: We shouldn't do `raise NotImplementedError` in the protocols. Protocols aren't actually inherited in the way that we're used to. Direct inheritance of protocols like
```
class TheImplementation(TheProtocol):
  ...
```
is actually discouraged. Weird things start happening when you do. Thus the default implementation in the protocol will not be called. We don't want to give people a false sense of security that a `NotImplementedError` will be raised, so we shouldn't make it look like it will.

dbt_semantic_interfaces/parsing/where_filter/where_filter_dimension.py
```diff
@@ -0,0 +1,68 @@
...
+        """The time granularity."""
+        self.time_granularity = TimeGranularity(time_granularity)
+        self._time_dimension_call_parameter_sets.append(
+            ParameterSetFactory.create_time_dimension(self.name, time_granularity, self.entity_path)
+        )
+        return self
+
+    def alias(self, _alias: str) -> QueryInterfaceDimension:
+        """Renaming the column."""
+        raise NotImplementedError
```
QMalcolm, line 47: Generally I would discourage leaf class `NotImplementedError`s in a PR except when either
1. We're implementing a protocol from another library
2. We plan on fast following in another PR to remove the `NotImplementedError`

I think this falls into (1) given that it sounds like these protocols were lifted from MF? Additionally I'm guess we also plan to implement `alias` shortly to be ready for the 1.7 release of core? If that sounds correct, this is good for now 🙂 

I don't believe 1 applies because

dbt_semantic_interfaces/parsing/where_filter/where_filter_dimension.py
```diff
@@ -0,0 +1,68 @@
...
+    ):
+        self.name = name
+        self.entity_path = entity_path
+        self._time_dimension_call_parameter_sets = time_dimension_call_parameter_sets
+        self.time_granularity: Optional[TimeGranularity] = None
+
+    def grain(self, time_granularity: str) -> QueryInterfaceDimension:
+        """The time granularity."""
+        self.time_granularity = TimeGranularity(time_granularity)
+        self._time_dimension_call_parameter_sets.append(
```
QMalcolm, line 40: Let me know if I'm getting out pocket on this. I'm likely biased from my personal preference for functional programming, but I found chain of passing the pointer to `time_dimension_factory.time_dimension_call_parameter_sets` odd. Additionally, until I followed the chain I was  confused as to why in the `where_filter_parser` dimension_parameter_sets were getting skipped unless they _didn't_ have a `time_granularity` specified. That is to say it appeared that the dimension_parameter_sets with `time_granularity` specified were getting trashed. I understand how it works now, and it's elegant in an OOP way. I do worry it's more complex than it needs to be though and will be harder for those unfamiliar with the code to approach. Additionally, thinking more about it, a question in my mind came up of what would happen if someone did the following
```python
{{ Dimension('<some_dimension>').grain('day').grain('week') }}
```
This chain wouldn't break on parsing and it would add the dimension to the `time_dimension_call_parameter_sets` twice, once with granularity `DAY` and once with granularity `WEEK`. Maybe that is what we want, but I don't think we do? My assumption here is that we'd actually just want it added once with granularity `WEEK`.

A way simplify things and to solve the duplication problem would be to not have `WhereFilterDimension`s directly modify the `time _dimension_call_parameter_sets`. Instead when `grain` is called it could just generate a `TimeDimensionCallParameterSet` and store it on a self property (maybe something like `self.time_dimension_call_parameter_set` which is defaulted to `None` on instantiation of the `WhereFilterDimsnion`. Then in ` parse_call_parameter_sets` when we iterate over `dimension_factory.created` we'd separate the created parameter sets into two lists, one of `DimensionCallParameterSet`s and the other of `TimeDimensionCallParameterSet`. The latter of which would be joined to the `time_dimension_factory.created` in the return statement.

DevonFulcher: > I was confused as to why in the where_filter_parser dimension_parameter_sets were getting skipped unless they didn't have a time_granularity specified

I will add comments to this section and look for other opportunities to add comments.

> {{ Dimension('<some_dimension>').grain('day').grain('week') }}

Good call. I hadn't considered that edge case.

> A way simplify things and to solve the duplication problem...

Good idea. I had implemented something similar to this, but I wasn't sure if the order of `time_dimension_call_parameter_sets` mattered, so I refactored to this current approach. If we did what you are suggesting, then the `TimeDimensionCallParameterSet`s that are generated from this syntax `Dimension('metric_time').grain('day')` would always be appended after the ones that are generated from the `TimeDimension` syntax. Paul mentioned that the order doesn't matter (but I had already completed the refactor), so I think I will go ahead and implement what you have suggested.

dbt_semantic_interfaces/parsing/where_filter/where_filter_parser.py
```diff
@@ -0,0 +1,55 @@
...
+
+        dimension_parameter_sets = []
+        for dimension in dimension_factory.created:
+            if not dimension.time_granularity:
+                param_set = ParameterSetFactory.create_dimension(dimension.name, dimension.entity_path)
+                dimension_parameter_sets.append(param_set)
+
+        return FilterCallParameterSets(
+            dimension_call_parameter_sets=tuple(dimension_parameter_sets),
+            time_dimension_call_parameter_sets=tuple(time_dimension_factory.time_dimension_call_parameter_sets),
```
tlento, line 53: Will this reference passing and hidden update extensions be addressed by the approach @QMalcolm suggests? I think it will. If so that's pretty nice, because this is confusing - it looks as if we skip any time dimensions defined through the Dimension() syntax, but that isn't the case.

tlento (commented):
Nice! I love the consolidation.

attention @plypaul this seems relevant to your work on saved queries.

dbt_semantic_interfaces/parsing/where_filter/where_filter_entity.py
```diff
@@ -0,0 +1,39 @@
...
+)
+from dbt_semantic_interfaces.parsing.where_filter.query_interface import (
+    QueryInterfaceEntity,
+    QueryInterfaceEntityFactory,
+)
+from dbt_semantic_interfaces.protocols.protocol_hint import ProtocolHint
+
+
+class EntityStub(ProtocolHint[QueryInterfaceEntity]):
+    """An Entity implementation that does nothing."""
```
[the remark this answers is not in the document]
DevonFulcher: Good call.

dbt_semantic_interfaces/parsing/where_filter/where_filter_parser.py
DevonFulcher: I think I have fixed your concern, but please let me know if you prefer a different approach. I don't think there is a way to implement this without the mutable `WhereFilterTimeDimensionFactory.time_dimension_call_parameter_sets`, `WhereFilterEntityFactory.entity_call_parameter_sets`, & `WhereFilterDimensionFactory.created` properties given the protocols that we want to implement.

dbt_semantic_interfaces/parsing/where_filter/where_filter_dimension.py
DevonFulcher: I believe I have addressed all of your concerns. Let me know if you would prefer a different approach.

QMalcolm (approved):
Thank you for this work as a whole and the updates! Looks good to me 🙂

dbt_semantic_interfaces/parsing/where_filter/where_filter_entity.py
```diff
@@ -16,7 +16,7 @@
 
 
 class EntityStub(ProtocolHint[QueryInterfaceEntity]):
-    """An Entity implementation that does nothing to satisfy the protocol.
+    """An Entity implementation that just satisfies the protocol.
```
tlento, line 19: Nice. 😁

tlento (commented):
Looks great, I think the new logic is easier to follow too!

PR #154, opened 2023-09-21
Support for descending & date_part in query interface

DevonFulcher:
This PR adds support for date_part and descending in the query interface. I decided to raise an error for `date_part` for now. We will have to revisit that.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/parsing/where_filter/where_filter_dimension.py
+++ b/dbt_semantic_interfaces/parsing/where_filter/where_filter_dimension.py
@@ -4,6 +4,7 @@ from typing import List, Optional, Sequence
 
 from typing_extensions import override
 
+from dbt_semantic_interfaces.errors import InvalidQuerySyntax
 from dbt_semantic_interfaces.protocols.protocol_hint import ProtocolHint
 from dbt_semantic_interfaces.protocols.query_interface import (
     QueryInterfaceDimension,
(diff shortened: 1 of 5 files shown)


dbt_semantic_interfaces/parsing/where_filter/where_filter_dimension.py
```diff
@@ -32,6 +33,14 @@ def grain(self, time_granularity: str) -> QueryInterfaceDimension:
...
         return self
 
+    def descending(self, _is_descending: bool) -> QueryInterfaceDimension:
+        """Set the sort order for order-by."""
+        raise InvalidQuerySyntax("descending is invalid in the where parameter and filter spec")
+
+    def date_part(self, _date_part: str) -> QueryInterfaceDimension:
+        """Date part to extract from the dimension."""
+        raise InvalidQuerySyntax("date_part isn't currently supported in the where parameter and filter spec")
+
```
tlento, line 43: I saw this in the other implementations as well and I wonder if there's a way for us to structure the protocols so that things are just typed, but I don't know if that's worth the complexity.

tlento: Related question: in practice, will anybody other than us (or another implementer of a MetricFlow query interface) ever initialize one of these things directly? I don't think the WhereFilterDimension is a thing any end user will ever interact with except via our parser, is that correct?

dbt_semantic_interfaces/protocols/query_interface.py
```diff
@@ -42,6 +61,8 @@ def create(
         self,
         time_dimension_name: str,
         time_granularity_name: str,
+        descending: Optional[bool] = None,
+        date_part_name: Optional[str] = None,
```
tlento, line 65: Should these also be added to the QueryInterfaceDimensionFactory.create() method?

tlento (approved):
Makes sense to me! We'll see if adjusting protocols makes sense as we get more of these implemented. For now being able to throw the errors from the relevant class, instead of some central handler, seems like a much bigger advantage than potential type safety, especially given how we expect these to be used.

dbt_semantic_interfaces/protocols/query_interface.py
tlento: Oops, I somehow forgot about this between writing it and submitting the review. If this was an oversight please fix before merge!

dbt_semantic_interfaces/parsing/where_filter/where_filter_dimension.py
DevonFulcher: > structure the protocols so that things are just typed

I believe this is typed correctly. Users of the JDBC interface don't get the benefits of an LSP, so they are reliant on error messages to help them write correct code.

> will anybody other than us (or another implementer of a MetricFlow query interface) ever initialize one of these things directly?

Right, I think this is just an internal class.

PR #156, opened 2023-09-26
Fix TimeDimension parameter ordering

DevonFulcher:
### Description

This PR fixes the ordering of the TimeDimension parameters so that entity_path is in the same position as it has been, right after time_granularity_name. This will prevent errors if a user attempts to use positional parameters to pass in entity_path, especially in the Jinja sandbox.

### Checklist


QMalcolm (approved):
Good catch, looks good to me!

PR #158, opened 2023-09-27
Support `label` attribute

### LINKED ISSUE #143: Add optional `label` attribute protocols
[dbt-core](https://github.com/dbt-labs/dbt-core) currently allows for specifying an optional `label` property  on [metric nodes](https://github.com/dbt-labs/dbt-core/blob/b39eeb328c9a245f21b54fb48f2761ad03d4d6ad/core/dbt/contracts/graph/nodes.py#L1463). However we don't currently specify this on the [metric protocol](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/ef8a893cc6e65f695db46356b8471a039246456b/dbt_semantic_interfaces/protocols/metric.py#L147). MetricFlow would like to begin using this property, and would like it to exist on more objects. 

We should add an optional `label` property to the following protocols:

- [Metric](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/ef8a893cc6e65f695db46356b8471a039246456b/dbt_semantic_interfaces/protocols/metric.py#L147)
- [SemanticModel](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/ef8a893cc6e65f695db46356b8471a039246456b/dbt_semantic_interfaces/protocols/semantic_model.py#L53)
- [Dimension](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/ef8a893cc6e65f695db46356b8471a039246456b/dbt_semantic_interfaces/protocols/dimension.py#L49)
- [Measure](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/ef8a893cc6e65f695db46356b8471a039246456b/dbt_semantic_interfaces/protocols/measure.py#L48)

QMalcolm:
Resolves #143

### Description

In this PR we
1. Add optional `label` attribute to the following protocols:
    - Metric
    - SavedQuery
    - SemanticModel
    - Dimension
    - Entity
    - Measure

2. Begin supporting optional `label` attribute on the corresponding pydantic implementations and parsing
3. Add validations which ensure
    - metric labels are unique to other metrics
    - saved query labels are unique to other saved queries
    - semantic model labels are unique to other semantic models
    - dimension labels on a semantic model are unique to other dimensions on the same semantic model
    - entity labels on a semantic model are unique to other entities on the same semantic model
    - measure labels on a semantic model are unique to other measures on the same semantic model
    - entities with the same name across semantic models have the same label (or are `None`)

### Checklist


tests/parsing/test_semantic_model_parsing.py
```diff
@@ -132,6 +186,38 @@ def test_semantic_model_entity_metadata_parsing() -> None:
     assert entity.metadata.file_slice.content == expected_metadata_content
 
 
+def test_base_semantic_model_measure_parsing() -> None:
+    """Test parsing base attributes of PydanticMeasure object."""
+    description = "Test sematic_model measure description"
```
tlento, line 191: nit: semantic_model, not sematic_model

tests/parsing/test_semantic_model_parsing.py
```diff
@@ -15,6 +15,31 @@
 )
 
 
+def test_base_semantic_model_parsing() -> None:
+    """Test parsing base attributes of PydanticSemanticModel object."""
+    description = "Test sematic_model description"
```
tlento, line 20: sematic_model's everywhere.....

tests/parsing/test_semantic_model_parsing.py
```diff
@@ -334,6 +420,37 @@ def test_semantic_model_primary_time_dimension_parsing() -> None:
     assert dimension.type_params is not None
 
 
+def test_base_semantic_model_dimension_parsing() -> None:
+    """Test parsing base attributes of PydanticDimension object."""
+    description = "Test sematic_model dimension description"
```
tlento, line 425: This is the cutest copy/paste gif I could find.....

cat on copier

tests/test_implements_satisfy_protocols.py
```diff
@@ -152,27 +155,9 @@ class RuntimeCheckableDimension(DimensionProtocol, Protocol):
...
-    # raising an exception during the assertion.
-    # of
-    # categorical_dim = PydanticDimension(
-    #     name="test_categorical_dim",
-    #     type=DimensionType.CATEGORICAL,
-    # )
-    # assert isinstance(categorical_dim, RuntimeCheckableDimension)
+@given(builds(PydanticDimension))
+def test_dimension_protocol(dimesnion: PydanticDimension) -> None:  # noqa: D
+    assert isinstance(dimesnion, RuntimeCheckableDimension)
```
tlento, line 160: How is this not subject to the same problem with categorical dimension types as the old test?

tests/test_implements_satisfy_protocols.py
```diff
@@ -1,45 +1,56 @@
...
+    PydanticMetric,
+    type=just(MetricType.SIMPLE),
+    type_params=builds(PydanticMetricTypeParams, measure=builds(PydanticMetricInputMeasure)),
+)
+
+SEMANTIC_MODEL_STRATEGY = builds(
+    PydanticSemanticModel,
+    dimensions=lists(builds(PydanticDimension)),
+    entities=lists(builds(PydanticEntity)),
+    measures=lists(builds(PydanticMeasure)),
```
tlento, line 52: Oh neat, it infers values based on type hints, I guess recursively all the way down.

dbt_semantic_interfaces/validations/labels.py
```diff
@@ -0,0 +1,43 @@
+import logging
```
tlento, line 1: I got to here..... will look over validations tomorrow!

tlento (commented):
Not done but I figured you might want to look at the cat gif earlier....

tests/test_implements_satisfy_protocols.py
QMalcolm: It does it pretty well! Although it sometimes jokes on nested lists of structured objects and just does an empty list instead. Hence why I explicitly tell it to generate lists of dimensions/entities/measures

QMalcolm:
Gonna do commits to address comments. I'll fix them up into the relevant commits before merging, but after reviewing is finished 🙂

tests/parsing/test_semantic_model_parsing.py
QMalcolm: Not marking comment as resolved because cat gif 🐈

dbt_semantic_interfaces/validations/labels.py
```diff
@@ -41,3 +42,103 @@ def validate_manifest(semantic_manifest: SemanticManifestT) -> Sequence[Validati
...
+        for dimension in semantic_model.dimensions:
+            if dimension.label is not None:
+                label_counts[dimension.label] = label_counts[dimension.label] + 1
+
+        for label, count in label_counts.items():
+            if count > 1:
+                issues.append(
+                    ValidationError(
+                        context=FileContext.from_metadata(semantic_model.metadata),
+                        message=f"Dimension labels must be unique within a semantic model. The label `{label}` was "
```
tlento, line 83: Do we want them to be globally unique for dimensions? I can see people setting `user__country` and `sales_rep__country` to both be `Country`, which isn't great.

tlento (commented):
Looks good to me, I have one open question about the dimension uniqueness rule.

tests/test_implements_satisfy_protocols.py
QMalcolm: Thats a good question. So I ran the test a few different ways after seeing your comment in this review.

1.
```python
import pdb
@given(builds(PydanticDimension))
def test_dimension_protocol(dimension: PydanticDimension) -> None:  # noqa: D
    if dimension.type == DimensionType.CATEGORICAL:
        pdb.set_trace()
    assert isinstance(dimension, RuntimeCheckableDimension)
```

2.
```python
import pdb
@given(builds(PydanticDimension))
def test_dimension_protocol(dimension: PydanticDimension) -> None:  # noqa: D
    if dimension.type == DimensionType.TIME:
        pdb.set_trace()
    assert isinstance(dimension, RuntimeCheckableDimension)
```

3.
```python
import pdb
@given(builds(PydanticDimension))
def test_dimension_protocol(dimension: PydanticDimension) -> None:  # noqa: D
    if dimension.type_params is not None:
        pdb.set_trace()
    assert isinstance(dimension, RuntimeCheckableDimension)
```

Variation (1) and (2) hit the break point when the test was ran, but (3) never was. Thus it appears the type_params aren't getting automatically generated by hypothesis. I'm gonna do some more investigating and see if this is happening to _all nested objects_. If so I'll have to build better strategies. I haven't found a setting in hypothesis to handle this better by default. It looks like we might have be able to [register strategies](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.register_type_strategy) for custom types, which is maybe what we have to do?

tests/test_implements_satisfy_protocols.py
QMalcolm: It also might be choking on nested objects in general. Or maybe it's only nested optional objects 🤔

dbt_semantic_interfaces/validations/labels.py
QMalcolm: For those following along, we discussed this in a call. The decision for now was to not require labels for dimensions or measures to be globally unique. If people end up consistently running into disambiguation problems, we'll look into being stricter

tests/test_implements_satisfy_protocols.py
QMalcolm: Digging in further, this might be a pydantic class defaulting issue https://github.com/HypothesisWorks/hypothesis/issues/3218. Which makes sense why I was seeing it on some things, but not on others.

tlento (approved):
Per discussion today the current validation looks good!

tests/test_implements_satisfy_protocols.py
QMalcolm: So I improved our strategies, but that was a separate fix. Why the tests no longer have the issue with categorical dimensions is we removed the assertion causing the issue in this commit https://github.com/dbt-labs/dbt-semantic-interfaces/commit/83cc44d2d896892e756cfd6a17f5bbe0456821a4

QMalcolm:
Rebasing to fixup commit [9373115](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/158/commits/93731159a66624a233a98daf7f1b3b40572069d8) into [59357a1](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/158/commits/59357a1efd52c5fdedd8206bf318eaf0fafff5fe)

PR #159, opened 2023-09-27
Add null value coalescing options to `MetricInputMeasure` protocol

### LINKED ISSUE #142: Add configurations for null value handing in metrics
A bug was raised in MetricFlow recently (https://github.com/dbt-labs/metricflow/issues/764) outlining how null values are not handled properly in some cases. The proposed solution (https://github.com/dbt-labs/metricflow/issues/759) requires changes to the `MetricInputMeasure` protocol used by `MetricTypeParams` protocol on `Metric` protocol. Specifically two new optional properties need to be added

- `join_to_timespine` which should have a boolean value
- `fill_nulls_with` which should optionally have a integer value

In the pydantic object implementations
- `join_to_timespine` should default to false if not specified during parsing / deserialization
- `fill_nulls_with` should default to None if not specified during parsing / deserialization

QMalcolm:
Resolves #142

### Description

Adds `join_to_timespine` and `fill_nulls_with` properties to `MetricInputMeasure` protocol

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/parsing/schemas.py
+++ b/dbt_semantic_interfaces/parsing/schemas.py
@@ -46,6 +46,8 @@ metric_input_measure_schema = {
                 "name": {"type": "string"},
                 "filter": {"type": "string"},
                 "alias": {"type": "string"},
+                "join_to_timespine": {"type": "boolean"},
+                "fill_nulls_with": {"type": "integer"},
             },
             "additionalProperties": False,
         },
--- a/dbt_semantic_interfaces/protocols/metric.py
+++ b/dbt_semantic_interfaces/protocols/metric.py
@@ -43,6 +43,18 @@ class MetricInputMeasure(Protocol):
         """Property accessor to get the MeasureReference with the aliased name, if appropriate."""
         ...
 
+    @property
+    @abstractmethod
+    def join_to_timespine(self) -> bool:
+        """If the measure should be joined to the timespine."""
+        pass
+
+    @property
+    @abstractmethod
+    def fill_nulls_with(self) -> Optional[int]:
+        """What null values should be filled with if set."""
+        pass
+
 
 class MetricTimeWindow(Protocol):
     """Describes the window of time the metric should be accumulated over, e.g., '1 day', '2 weeks', etc."""
--- a/.changes/unreleased/Features-20230926-182305.yaml
+++ b/.changes/unreleased/Features-20230926-182305.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add options to protocol for null value coalescing
+time: 2023-09-26T18:23:05.191433-07:00
+custom:
+  Author: QMalcolm
+  Issue: "142"
--- a/dbt_semantic_interfaces/implementations/metric.py
+++ b/dbt_semantic_interfaces/implementations/metric.py
@@ -30,6 +30,8 @@ class PydanticMetricInputMeasure(PydanticCustomInputParser, HashableBaseModel):
     name: str
     filter: Optional[PydanticWhereFilter]
     alias: Optional[str]
+    join_to_timespine: bool = False
+    fill_nulls_with: Optional[int] = None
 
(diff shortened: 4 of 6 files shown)


dbt_semantic_interfaces/protocols/metric.py
```diff
@@ -43,6 +43,18 @@ def post_aggregation_measure_reference(self) -> MeasureReference:
...
 
+    @property
+    @abstractmethod
+    def join_to_timespine(self) -> bool:
+        """Where the measure should be joined to the timespine."""
+        pass
+
+    @property
+    @abstractmethod
+    def fill_nulls_with(self) -> Optional[int]:
```
tlento, line 54: I think this should eventually be expanded to allow floats, since measures can be averages and things like that. Oddly enough, the `float` type also allows int types, but the reverse is not true: https://peps.python.org/pep-0484/#the-numeric-tower

However, it's annoying, because we don't want to render SQL like `COALESCE(average_revenue, 0.0000000000017)` or whatever.

Perhaps we should return `Optional[Union[int, float]]` and then explicitly check at the callsite, since how we format the value into the SQL might be different depending on whether it's an int or a float.

We can leave this as-is for now, since widening the type is backwards compatible and by far the most common fallback value will be 0, but if we already know int | float is what we want we might as well get that rolling now.

dbt_semantic_interfaces/parsing/schemas.py
```diff
@@ -46,6 +46,8 @@
                 "name": {"type": "string"},
                 "filter": {"type": "string"},
                 "alias": {"type": "string"},
+                "join_to_timespine": {"type": "boolean"},
+                "fill_nulls_with": {"type": "integer"},
```
tlento, line 50: If we change the type this, of course, has to change along with it.

dbt_semantic_interfaces/protocols/metric.py
```diff
@@ -43,6 +43,18 @@ def post_aggregation_measure_reference(self) -> MeasureReference:
         """Property accessor to get the MeasureReference with the aliased name, if appropriate."""
         ...
 
+    @property
+    @abstractmethod
+    def join_to_timespine(self) -> bool:
+        """Where the measure should be joined to the timespine."""
```
tlento, line 49: nit:
```suggestion
        """If the measure should be joined to the timespine."""
```

tlento (approved):
I'm on board with this as-is, although maybe we need a float (which includes int) type instead of int?

dbt_semantic_interfaces/protocols/metric.py
QMalcolm: I'm gonna update it now to support both, that is `Optional[Union[int, float]]` so that we have less work to do in the future. I think this means I'll also have to update it in core, but eh oh well.

QMalcolm: So I looked into this aaaaand there's some wonkiness. JsonSchema has two types for numerics, `number` and `integer`. The interesting part is `number` will handle both integers and floats, whereas `integer` only handle integers. Thus making it seem like the solution on the JsonSchema side would be to specify `fill_nulls_with` as `number`. The problem though is on the python/pydantic side. If you specify `Union[float, int]` it is always parsed as a float. If you specify it as `Union[int, float]`, it'll always parse as an int. I'm not entirely sure if this is a python issue or a pydantic issue. There are some possible ways around it, but it'll require a fair amount of extra work. The truth is most people will be setting this to 0, or possibly 1. If someone ends up needing this to be a float, we'll address it then.

QMalcolm:
About to fixup https://github.com/dbt-labs/dbt-semantic-interfaces/pull/159/commits/d0f0152e8a9439780f33406533fe9487364008c2 into https://github.com/dbt-labs/dbt-semantic-interfaces/pull/159/commits/47ea1fb3d888f4010d0209662c52dd01fed65084 and then force push it up to this branch

PR #161, opened 2023-09-29
[BACKPORT 0.2.latest] Support for null value coalescing

### LINKED ISSUE #142: Add configurations for null value handing in metrics

QMalcolm:
Resolves #142 

### Description

The lack of null value coalescing is a bug in 0.2.x because it makes some metrics incorrect. We're backporting the addition of these options to correct that. This is specifically a backport of #159 

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230926-182305.yaml
+++ b/.changes/unreleased/Features-20230926-182305.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add options to protocol for null value coalescing
+time: 2023-09-26T18:23:05.191433-07:00
+custom:
+  Author: QMalcolm
+  Issue: "142"
--- a/dbt_semantic_interfaces/implementations/metric.py
+++ b/dbt_semantic_interfaces/implementations/metric.py
@@ -30,6 +30,8 @@ class PydanticMetricInputMeasure(PydanticCustomInputParser, HashableBaseModel):
     name: str
     filter: Optional[PydanticWhereFilter]
     alias: Optional[str]
+    join_to_timespine: bool = False
+    fill_nulls_with: Optional[int] = None
 
     @classmethod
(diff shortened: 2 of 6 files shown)


tlento (approved):
Seems reasonable! I assume we won't be able to deploy this until core 1.6.x is updated, correct? Or is this also backwards compatible with existing installations? Like the parser should be fine until someone tries to actually set one of these parameters, but otherwise the defaults take over? When we add support for this to MetricFlow we'll need to ensure the MF dependency specifies the associated 0.2.x as the minimum version.

QMalcolm:
> Seems reasonable! I assume we won't be able to deploy this until core 1.6.x is updated, correct? Or is this also backwards compatible with existing installations? Like the parser should be fine until someone tries to actually set one of these parameters, but otherwise the defaults take over? When we add support for this to MetricFlow we'll need to ensure the MF dependency specifies the associated 0.2.x as the minimum version.

So from the core side, these options don't exist in 1.6 yet. That's okay because they are optional / will get defaulted. The parser will be fine unless someone tries to specify one of these. With that said the plan in core was to backport these options to 1.6. We've done the work to support them in the upcoming 1.7 release, and I was planning on backporting it to 1.6 later this week.

tlento:
Sounds good, merge when convenient!

PR #162, opened 2023-09-29
Backport: Add support for `Dimension(...).grain(...)` in where filter

DevonFulcher:
Resolving: SL-631

### Description

Due to MetricFlow's reliance on version 0.2.x of DSI, the updates to support `Dimension(...).grain(...)` need to be made to this branch as well. I just combined the work from these PRs into this one https://github.com/dbt-labs/dbt-semantic-interfaces/pull/156, https://github.com/dbt-labs/dbt-semantic-interfaces/pull/154, https://github.com/dbt-labs/dbt-semantic-interfaces/pull/152

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20230929-123932.yaml
+++ b/.changes/unreleased/Features-20230929-123932.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: 'Backport: Add support for Dimension(...).grain(...) syntax in where parameter'
+time: 2023-09-29T12:39:32.834352-05:00
+custom:
+  Author: DevonFulcher
+  Issue: None
--- a/dbt_semantic_interfaces/call_parameter_sets.py
+++ b/dbt_semantic_interfaces/call_parameter_sets.py
@@ -13,7 +13,7 @@ from dbt_semantic_interfaces.type_enums import TimeGranularity
 
 @dataclass(frozen=True)
 class DimensionCallParameterSet:
-    """When 'dimension(...)' is used in the Jinja template of the where filter, the parameters to that call."""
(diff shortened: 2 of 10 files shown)


tlento (commented):
This all seems fine to me, given my understanding of the changes and how the backport deployment works with core 1.6.x,  but I'd like @QMalcolm to take a look to confirm this is backwards compatible with existing 1.6 installations. In particular, I think there are some changes to the where_filter_parser along with the file move here.

QMalcolm (commented):
I unfortunately don't think we can back port this. Core imports the WhereFilterParser, and this changes that import path. Core's import of DSI is `~=0.2.0` in dbt-core `1.6.0` through `1.6.5`, if we backported and released this, all fresh installs of those versions would break.

tlento:
> I unfortunately don't think we can back port this. Core imports the WhereFilterParser, and this changes that import path. Core's import of DSI is ~=0.2.0 in dbt-core 1.6.0 through 1.6.5, if we backported and released this, all fresh installs of those versions would break.

Is that the only issue? If so, can we add an import alias to the backport then? We should be able to put in a module in the original location that imports this WhereFilterParser (or defines a WhereFilterParser class that wraps it, if there are API changes that need to be managed).

QMalcolm (changes requested):
@DevonFulcher brought up the idea of moving the `where_filter_parser.py` file back out for the backport in slack, also @tlento brought it in an above comment. Combing through these changes, and imports of `dbt-semantic-interfaces` in `dbt-core`, I think we can get away with this by doing so.

QMalcolm (approved):

tlento:
Yay! 🚢

PR #170, opened 2023-10-10
Throw exceptions on invalid entity names in filter and group by expressions

tlento:
The previous logic was simply calling an error string formatting function,
and one which would return an incorrect error message at that. This raises
an exception with an entity-specific name formatting error message.

PR #171, opened 2023-10-10
Allow specification of filters as either lists or strings

tlento:
In order to support more advanced forms of filter rendering for things like
predicate pushdown (see https://github.com/dbt-labs/metricflow/issues/747),
we need a way for users to indicate that a given subset of a complex filter
expression can be treated as an independent part of an intersection of
multiple filters.

Providing a construct that can accept, at the config level, a list as well as a string
gives us the flexibility we need to operate on sub-filters as independent entities
while maintaining backwards compatibility with the existing bare string filter
expressions in the wild today.

This PR does all of the necessary work for enabling filter lists in the interface
itself, by adding a filter container protocol class and corresponding Pydantic
implementation, along with test cases and json schema updates to allow
either list or string input in our sample YAML spec.

This does not appear to be backport-eligible, as the output property change
will not be backwards compatible in dbt-core 1.6 even though the Pydantic
object stack that MetricFlow uses internally should cover the discrepancy
in the query runtime.

### DIFF
--- a/.changes/unreleased/Features-20231009-210737.yaml
+++ b/.changes/unreleased/Features-20231009-210737.yaml
@@ -1,0 +1,7 @@
+kind: Features
+body: Allow metric filters and saved query where properties to accept lists of filter
+  expressions
+time: 2023-10-09T21:07:37.978465-07:00
+custom:
+  Author: tlento
+  Issue: "147"
--- a/dbt_semantic_interfaces/implementations/filters/where_filter.py
+++ b/dbt_semantic_interfaces/implementations/filters/where_filter.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
-from typing import Callable, Generator, List
+from typing import Callable, Generator, List, Tuple
 
 from typing_extensions import Self
 
-from dbt_semantic_interfaces.call_parameter_sets import FilterCallParameterSets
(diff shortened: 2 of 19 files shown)


tlento:
Current dependencies on/for this PR:
* main
  * **PR #170** <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-semantic-interfaces/170" target="_blank">Graphite</a> 
    * **PR #171** <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-semantic-interfaces/171" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/dbt-semantic-interfaces/171?utm_source=stack-comment).

QMalcolm (approved):
This is incredibly well put together and detailed. I want my code to look like this. I found no issues or nits on my first or second pass, not even typos. I think the level of detail this goes into to massage things to keep things backwards compatible will make it much easier for me on the core side of things. Ship it 🚀

tlento:
Glad you appreciate the PR construction effort! The parsing changes were hard for me to reason about so I kind of had to write it all down first. I did that in the docblock just so I wouldn't lose it, which proved to be a good call, made it easy to clean up into documentation.

Let's just not talk about how many times I screwed up and had to rewrite my git history along the way....

PR #183, opened 2023-10-23
Support for date_part for Dimension & TimeDimension

DevonFulcher:
### Description

This PR adds support for the date_part for Dimension & TimeDimension. This is helpful for our Tableau integration. A `time_dimension_call_parameter_set` is now created if `Dimension(...).grain(...)` or `Dimension(...).date_part(...)` is used or both.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
+++ b/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
@@ -17,6 +17,7 @@ from dbt_semantic_interfaces.references import (
     TimeDimensionReference,
 )
 from dbt_semantic_interfaces.type_enums import TimeGranularity
+from dbt_semantic_interfaces.type_enums.date_part import DatePart
 
 
 class ParameterSetFactory:
@@ -31,7 +32,10 @@ class ParameterSetFactory:
 
     @staticmethod
     def create_time_dimension(
-        time_dimension_name: str, time_granularity_name: Optional[str] = None, entity_path: Sequence[str] = ()
+        time_dimension_name: str,
+        time_granularity_name: Optional[str] = None,
+        entity_path: Sequence[str] = (),
+        date_part_name: Optional[str] = None,
     ) -> TimeDimensionCallParameterSet:
         """Gets called by Jinja when rendering {{ TimeDimension(...) }}."""
         group_by_item_name = DunderedNameFormatter.parse_name(time_dimension_name)
@@ -57,6 +61,7 @@ class ParameterSetFactory:
                 tuple(EntityReference(element_name=arg) for arg in entity_path) + group_by_item_name.entity_links
             ),
             time_granularity=TimeGranularity(time_granularity_name) if time_granularity_name is not None else None,
+            date_part=DatePart(date_part_name.lower()) if date_part_name else None,
         )
 
     @staticmethod
--- a/dbt_semantic_interfaces/type_enums/date_part.py
+++ b/dbt_semantic_interfaces/type_enums/date_part.py
@@ -1,0 +1,43 @@
+from enum import Enum
+from typing import List
+
+from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
+from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
+
+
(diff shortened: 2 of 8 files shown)


dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
```diff
@@ -31,7 +32,10 @@ def _exception_message_for_incorrect_format(element_name: str) -> str:
 
     @staticmethod
     def create_time_dimension(
-        time_dimension_name: str, time_granularity_name: Optional[str] = None, entity_path: Sequence[str] = ()
+        time_dimension_name: str,
+        time_granularity_name: Optional[str] = None,
+        entity_path: Sequence[str] = (),
```
courtneyholcomb, line 37: Do we allow keyword arguments for TimeDimensions? I've gotten errors on that before but not sure if it's changed! I ask because as is, you would be forced to pass something in the `entity_path` field in order to pass a `date_part_name`. So it would be good to fix that! And/or change the order of params here so that date_part comes first (but that might be a breaking change)

tests/parsing/test_where_filter_parsing.py
```diff
@@ -135,3 +139,17 @@ def test_where_filter_intersection_from_partially_deserialized_list_of_strings()
     parsed_model = ModelWithWhereFilterIntersection.parse_obj(obj)
 
     assert parsed_model.where_filter == expected_parsed_output
+
+
+def test_time_dimension_date_part() -> None:  # noqa
```
courtneyholcomb, line 144: Should we add tests without all the params?

courtneyholcomb (approved):

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
DevonFulcher: Ya, we can't change the order of the params. I added support for kwargs here https://linear.app/dbt-labs/issue/SL-1100/multiple-updates-to-query-interface

courtneyholcomb: Sweet!!

dbt_semantic_interfaces/type_enums/date_part.py
```diff
@@ -0,0 +1,44 @@
...
+class DatePart(Enum):
+    """Date parts able to be extracted from a time dimension.
+
+    TODO: add support for hour, minute, second once those granularities are available
+    """
+
+    YEAR = "year"
+    QUARTER = "quarter"
+    MONTH = "month"
+    WEEK = "week"
```
tlento, line 17: We no longer support WEEK, and I'll be merging that change to MetricFlow tonight or tomorrow. https://github.com/dbt-labs/metricflow/pull/812

tlento (approved):
Let's not add support for WEEK here, I'll just need to remove it again later. And we might want to update the docstring on the DatePart enum to indicate the reasoning. 

Otherwise, this looks good to me!

tests/parsing/test_where_filter_parsing.py
QMalcolm: ➕

dbt_semantic_interfaces/type_enums/date_part.py
QMalcolm: ➕

QMalcolm (commented):
This is looking good! Some other people left comments on some things to add / change, and I'd like to see those addressed before approval.

Separately, I noticed that the `issue` in the changie doc is listed as `None`. Can we start creating issues for this kind of work moving forward? We have a [nice template](https://github.com/dbt-labs/dbt-semantic-interfaces/issues/new?assignees=&labels=user+docs&projects=&template=implementation-ticket.yml&title=%5B%3Cproject%3E%5D+%3Ctitle%3E) for this kind of internally motivated work. Having the issue is helpful for when we generate the release notes with changie. Having the issue in advance gives us in core an idea of whats coming down the pipe.

Additionally, is this going to need to be backported and put into a 0.3.latest patch release? If so we need to plan this work on the core side. Again, having a DSI issue will help with this.

PR #187, opened 2023-10-24
Assorted Naming and Type Fixes

plypaul:
### Description

* Use nouns for fields - `fill_nulls_with` -> `null_fill_value`
* Update the type for the null fill value to allow for floats.
* Rename MetricInput.as_reference -> MetricInput.metric_reference
* `timespine` -> `time_spine` in field names since it's 2 words.

### Checklist


QMalcolm:
@plypaul this is currently marked as a draft. Are we trying to get this into into 0.4.0 of DSI?

plypaul:
@QMalcolm - Picking this back up again. What would be the best way forward on this?

dbt_semantic_interfaces/implementations/metric.py
```diff
@@ -28,10 +28,10 @@ class PydanticMetricInputMeasure(PydanticCustomInputParser, HashableBaseModel):
...
 
     name: str
-    filter: Optional[PydanticWhereFilterIntersection]
-    alias: Optional[str]
-    join_to_timespine: bool = False
-    fill_nulls_with: Optional[int] = None
+    filter: Optional[PydanticWhereFilterIntersection] = None
+    alias: Optional[str] = None
+    time_spine_join: bool = False
+    null_fill_value: Optional[float] = None
```
tlento, line 34: When we tried this float caused problems when the actual value is supposed to be an int.

In a perfect world we could do `int | float` but that doesn't appear to resolve correctly and we haven't figured out how to deal with it.

dbt_semantic_interfaces/protocols/metric.py
```diff
@@ -46,13 +46,13 @@ def post_aggregation_measure_reference(self) -> MeasureReference:
...
-    def join_to_timespine(self) -> bool:
-        """If the measure should be joined to the timespine."""
+    def time_spine_join(self) -> bool:
+        """If the measure should be joined to the time spine."""
         pass
 
     @property
     @abstractmethod
-    def fill_nulls_with(self) -> Optional[int]:
+    def null_fill_value(self) -> Optional[float]:
```
tlento, line 55: These are hard breaking changes on the dbt-core parser so we can't push these into dbt-semantic-interfaces or take dependencies on these names in MetricFlow if we want to be able to backport things.

It's fine to add duplicate fields with the new names, and once we get support for both broadly available across the releases we are committed to supporting we can gradually remove the old ones, but given that 1) these names were chosen within the context of existing dbt metrics nomenclature and 2) they're now out in the world I'm not sure this adjustment is worth the effort.

tlento (changes requested):
None of the original blockers on this PR have changed, but for whatever reason we didn't write them down, so now I'm writing them down.

QMalcolm (commented):
I just want to echo everything @tlento has already said. If we aren't trying to move forward with this renames/typings should we close this for the time being?

PR #194, opened 2023-10-26
Add typing-extensions>=4.4 requirement

Fatal1ty:
Resolves #193
Related to https://github.com/dbt-labs/dbt-core/issues/7828

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #199, opened 2023-10-26
[Backport] Update typing-extensions to require versions ~=4.4 (#194)

tlento:
Support for the `@override` keyword was added with typing-extensions 4.4,
but the previous version spec made it possible for end users to encounter
errors when attempting to use dbt or MetricFlow.


QMalcolm (approved):
Perfect! Thank you 🙂

PR #200, opened 2023-10-26
Prevent typing-extensions from auto-updating to 5.x

tlento:
The typing-extensions project announced adherence to semver with
4.0.0, which means, at least in theory, 5.0 might include
breaking changes.

Rather than risk user packages auto-updating to backwards-incompatible
versions, we pin this back to ~=4.0, which is equivalent to >=4.0, <5.


QMalcolm (approved):
Looks great! Thank you!

PR #203, opened 2023-10-31
Throw exceptions on invalid entity names in filter and group by expressions

tlento:
The previous logic was simply calling an error string formatting function,
and one which would return an incorrect error message at that. This raises
an exception with an entity-specific name formatting error message.


QMalcolm (approved):
Looks good, good catch!

PR #209, opened 2023-11-16
Add Conversion Metrics Properties

WilliamDee:
### Description
- Added protocols/implementations for conversion metrics properties 
- Added transformation step for input measures for conversion metrics
- Added parser updates to handle parsing conversion metrics properties

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/parsing/schemas.py
+++ b/dbt_semantic_interfaces/parsing/schemas.py
@@ -9,8 +9,11 @@ TRANSFORM_OBJECT_NAME_PATTERN = "(?!.*__).*^[a-z][a-z0-9_]*[a-z0-9]$"
 
 
 # Enums
-metric_types_enum_values = ["SIMPLE", "RATIO", "CUMULATIVE", "DERIVED"]
+metric_types_enum_values = ["SIMPLE", "RATIO", "CUMULATIVE", "DERIVED", "CONVERSION"]
 metric_types_enum_values += [x.lower() for x in metric_types_enum_values]
+
+calculation_types_enum_values = ["CONVERSIONS", "CONVERSION_RATE"]
+calculation_types_enum_values += [x.lower() for x in calculation_types_enum_values]
 
 entity_type_enum_values = ["PRIMARY", "UNIQUE", "FOREIGN", "NATURAL"]
 entity_type_enum_values += [x.lower() for x in entity_type_enum_values]
@@ -85,6 +88,20 @@ metric_input_schema = {
     "additionalProperties": False,
 }
 
+conversion_type_params_schema = {
+    "$id": "conversion_type_params_schema",
+    "type": "object",
+    "properties": {
+        "base_measure": {"$ref": "metric_input_measure_schema"},
+        "conversion_measure": {"$ref": "metric_input_measure_schema"},
+        "calculation": {"enum": calculation_types_enum_values},
+        "entity": {"type": "string"},
+        "window": {"type": "string"},
+    },
+    "additionalProperties": False,
+    "required": ["base_measure", "conversion_measure", "entity"],
+}
+
 metric_type_params_schema = {
     "$id": "metric_type_params",
     "type": "object",
@@ -99,6 +116,7 @@ metric_type_params_schema = {
             "type": "array",
             "items": {"$ref": "metric_input_schema"},
         },
+        "conversion_type_params": {"$ref": "conversion_type_params_schema"},
     },
     "additionalProperties": False,
 }
@@ -382,6 +400,7 @@ schema_store = {
     filter_schema["$id"]: filter_schema,
     metric_input_measure_schema["$id"]: metric_input_measure_schema,
     metric_type_params_schema["$id"]: metric_type_params_schema,
+    conversion_type_params_schema["$id"]: conversion_type_params_schema,
     entity_schema["$id"]: entity_schema,
     measure_schema["$id"]: measure_schema,
     dimension_schema["$id"]: dimension_schema,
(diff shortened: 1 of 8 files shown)


dbt_semantic_interfaces/protocols/metric.py
```diff
@@ -212,6 +212,11 @@ def metrics(self) -> Optional[Sequence[MetricInput]]:  # noqa: D
     def conversion_type_params(self) -> Optional[ConversionTypeParams]:  # noqa: D
         pass
 
+    @property
+    @abstractmethod
+    def conversion_type_params_or_error(self) -> ConversionTypeParams:  # noqa: D
```
tlento, line 240: Oof. I don't think we want this on the type params protocol.

If this is a real problem we can add a helper like `assert_and_return(input: Optional[T]) -> T` inside MetricFlow, or if we're doing this in a handful of places adding an `assert metric.type_params.conversion_type_params is not None` call before accessing the property. I'm pretty sure MyPy is currently smart enough to refine the return type on that property for us.

dbt_semantic_interfaces/protocols/metric.py
```diff
@@ -117,6 +117,23 @@ def post_aggregation_reference(self) -> MetricReference:
         pass
 
 
+class ConstantPropertyInput(Protocol):
+    """Provides the constant property values for conversion metrics.
+
+    The specified property will be a reference of a dimension/entity.
```
tlento, line 123: I don't think anybody other than us will know what a `constant property` is without a more explicit description. Maybe something like:

```
"""
Provides the constant property set for conversion metrics.

Constant properties are additional elements linking a base event to a conversion event. 
The specified properties will typically be a reference to a dimension or entity, and will be used
to join the base event to the final conversion event. Typical constant properties are things like
session keys (for services where conversions are measured within a user session), or secondary entities
(like a user/application pair for an app platform or a user/shop pair for a retail/online storefront platform).
"""
```

dbt_semantic_interfaces/protocols/metric.py
```diff
@@ -215,3 +270,9 @@ def metadata(self) -> Optional[Metadata]:  # noqa: D
     def label(self) -> Optional[str]:
         """Returns a string representing a human readable label for the metric."""
         pass
+
+    @property
+    @abstractmethod
+    def conversion_params(self) -> ConversionTypeParams:
```
tlento: I see this was removed in a subsequent commit and I think that's the right approach, we don't need this here.

tlento (commented):
I think there's a general issue of too many accessors, and I think we'll be better off starting with more streamlined protocols. In particular, I'd like to avoid the `get_property_or_error` accessor pattern you've introduced here.

Otherwise, this looks good. We can see if @QMalcolm has any added feedback while we clean up the accessors.

dbt_semantic_interfaces/protocols/metric.py
QMalcolm: I would not have known what they were without this context ➕

tests/test_implements_satisfy_protocols.py
```diff
@@ -200,6 +200,18 @@ def test_metric_protocol_derived(metric: PydanticMetric) -> None:  # noqa: D
     assert isinstance(metric, RuntimeCheckableMetric)
 
 
+@given(
+    builds(
+        PydanticMetric,
+        type=just(MetricType.CONVERSION),
+        type_params=builds(PydanticMetricTypeParams, metrics=lists(builds(PydanticMetricInput))),
```
QMalcolm, line 207: I was under the impression that `conversion` metrics don't use `metric.type_params.metrics` but instead use the measure inputs specified in `metric.type_params.conversion_type_params`. Is there a reason why we need to provide them for this test?

QMalcolm (commented):
This is looking really good! Nothing major from me. @tlento has raised some really good points which I'd also like to see addressed 🙂

dbt_semantic_interfaces/protocols/metric.py
QMalcolm: ➕ ➕ We've had this logic crop up before, and I'd prefer us to push it out of DSI. It seems unnecessary to have on the protocols, and the DSI provided implementations.

dbt_semantic_interfaces/protocols/metric.py
```diff
@@ -113,6 +117,52 @@ def post_aggregation_reference(self) -> MetricReference:
...
+    @abstractmethod
+    def base_measure_reference(self) -> MeasureReference:
+        """Return the measure reference associated with the base measure."""
+        pass
+
+    @property
+    @abstractmethod
+    def conversion_measure_reference(self) -> MeasureReference:
+        """Return the measure reference associated with the conversion measure."""
+        pass
```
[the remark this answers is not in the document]
QMalcolm: ➕➕ I agree. We want to keep the protocols slim as slim as reasonably possible. Everything we add as a property, an implementer of the protocol then has to implement. I don't think these provides great enough convenience to justify that burden. For instance, we'd have to implement these in core just to satisfy the protocol, when accessing the underlying `MeasureReference` as a pretty straightforward path already.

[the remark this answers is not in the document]
WilliamDee: sounds reasonable!

dbt_semantic_interfaces/protocols/metric.py
WilliamDee: yea it was just annoying having to assert, but makes sense that we sohuld keep it out of DSI. I'll probably do something similar to what Tom suggested in MF

dbt_semantic_interfaces/parsing/schemas.py
```diff
@@ -9,9 +9,12 @@
 
 
 # Enums
-metric_types_enum_values = ["SIMPLE", "RATIO", "CUMULATIVE", "DERIVED"]
+metric_types_enum_values = ["SIMPLE", "RATIO", "CUMULATIVE", "DERIVED", "CONVERSION"]
```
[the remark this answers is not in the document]
WilliamDee: good idea, could do this for all the other enum values as well!

tests/test_implements_satisfy_protocols.py
WilliamDee: whoops good catch, i forgot to update this. Updated now!

QMalcolm (approved):
Looks good to me! Thank you for doing this work 🙂

PR #214, opened 2023-11-21
Dev release bug fix: compatible granularities for date_part

courtneyholcomb:
### Checklist

### DIFF
--- a/dbt_semantic_interfaces/type_enums/date_part.py
+++ b/dbt_semantic_interfaces/type_enums/date_part.py
@@ -40,4 +40,4 @@ class DatePart(Enum):
     @property
     def compatible_granularities(self) -> List[TimeGranularity]:
         """Granularities that can be queried with this date part."""
-        return [granularity for granularity in TimeGranularity if granularity.to_int() >= self.to_int()]
+        return [granularity for granularity in TimeGranularity if granularity.to_int() <= self.to_int()]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.4.1"
+version = "0.4.2dev0"
 description = 'The shared semantic layer definitions that dbt-core and MetricFlow use'
 readme = "README.md"
 requires-python = ">=3.8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.4.2dev0"
+version = "0.4.2.dev0"
 description = 'The shared semantic layer definitions that dbt-core and MetricFlow use'
 readme = "README.md"
 requires-python = ">=3.8"


dbt_semantic_interfaces/type_enums/date_part.py
```diff
@@ -40,4 +40,4 @@ def to_int(self) -> int:
     @property
     def compatible_granularities(self) -> List[TimeGranularity]:
         """Granularities that can be queried with this date part."""
```
tlento, line 42: This is so confusing, can we explain it here?

We are switching this to allow TimeGranularity.DAY for DatePart.MONTH, correct? So it'd now look like:

```
SELECT date_part(month, metric_time__day), count(*)
FROM ...
GROUP BY date_part(month, metric_time__day)
```

Whereas before we only allowed something like this:

```
SELECT date_part(day, metric_time__month), count(*)
FROM ...
GROUP BY date_part(day, metric_time__month)
```

Ok, that makes sense to me now.

Maybe we can document this as:

```
We support querying granularities that are the same underlying granularity or coarser than the requested date_part. This is because extracting the "day" part from a monthly granularity will always produce a date_part value of `1`, which is probably not what the user wants.
```

tlento (approved):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.4.1"
+version = "0.4.2dev0"
```
tlento, line 3: Oh wait, this is wrong!

tlento (changes requested):
Oops, please fix the version number! Requesting changes to avoid unintentional reverts of @QMalcolm's changes. Sorry!

dbt_semantic_interfaces/type_enums/date_part.py
courtneyholcomb: @tlento this is just the dev release for a PR that already merged to main, you can see [here](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/212).
This is not actually a behavior change for queries, this was only used in an error message.

pyproject.toml
courtneyholcomb: 🤔 This is the version that @QMalcolm tole me to use!

QMalcolm: @tlento Courtney and I talked about this. The goal is to do a dev release so that it can be tested in MetricFlow without doing a production patch release. What we do is a `0.4.2.dev0`, and if all goes well we then follow up with a `0.4.2` release. I don't anticipate anything bad in doing this. The `~=` will automatically ignore versions ending in `.dev`, `rc`, `b`, and `a` so long as none of those specifications exist in the version specified on the right hand side of `~=` specification.

QMalcolm: Oh is this that it should be `0.4.2.dev0` instead of `0.4.2dev0`. That is likely on me. `.` are expected with `dev` versions but not `rc`, `b`, and `a` strangely 🤔

QMalcolm: I'll be out the rest of the week, so please move forward without me 🙂

tlento (approved):
Sorry I didn't read the target branch name, this is totally fine (although maybe .dev for the version number is better).

PR #227, opened 2023-12-13
fix: relax constraint on pydantic

bernardcooke53:
This change allows DSI to be installed in environments alongside either pydantic v1 or v2

~Resolves~ Doesn't _resolve_ #134, but does make it less impactful to end users

### Description

Pydantic retained the v1 codebase in a `v1` namespace when upgrading to v2, per their [Migration Guide](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features). Therefore it's possible to make minimal code changes while allowing DSI to co-exist alongside either Pydantic v1 or v2.

According to the discussion on #217, this is an issue that might take some time to puzzle through fully given the need to support both Pydantic v1 and v2. This PR would alleviate the constraint on Pydantic without requiring additional testing or impact on end-users, so hopefully also relieves some pressure on the maintainers with regards to timescales. It might (I haven't checked for certain) enable a more gradual upgrade path, module-by-module.

The downside being that I've excluded pydantic from mypy checking for missing imports - I can change the code so that it isn't globally ignored, but I figured this would be an issue everywhere.
I completely understand if this gets closed as "not the right approach" - just thought I'd add what I can to help 🙂 

### Checklist

### DIFF
--- a/.changes/unreleased/Dependencies-20231213-113016.yaml
+++ b/.changes/unreleased/Dependencies-20231213-113016.yaml
@@ -1,0 +1,7 @@
+kind: Dependencies
+body: Relax Pydantic dependency to allow DSI to be installed in environments alongside
+  Pydantic v1 or v2
+time: 2023-12-13T11:30:16.376387Z
+custom:
+  Author: bernardcooke53
+  PR: "227"
--- a/dbt_semantic_interfaces/dataclass_serialization.py
+++ b/dbt_semantic_interfaces/dataclass_serialization.py
@@ -19,8 +19,16 @@ from typing import (
     get_type_hints,
 )
 
-import pydantic
-from pydantic import BaseModel
+try:
+    import pydantic.v1 as pydantic
+except (ImportError, ModuleNotFoundError):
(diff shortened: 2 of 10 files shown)


esciara:
Thanks for trying out !

Carefull though: there are a lot of behavioural changes from pydantic v1 to v2. Notably around [required optional and nullable fields](https://docs.pydantic.dev/2.0/migration/#required-optional-and-nullable-fields). Wonder whether at the very least this should be applied...

Do these test pass with pydantic v2 as dependency?

bernardcooke53:
Oh - I should also add - because we always try to import from `pydantic.v1` first, we'll get the v1 objects, even if v2 is installed. Because we're catching `ModuleNotFoundError` instead of `AttributeError` there shouldn't be a risk that a typo (e.g. `from pydantic.v1 import spam`) will result in _accidentally_ trying to import that object from `pydantic` due to the try-except

bernardcooke53:
Think I'll close this as #238 supersedes it, and is cleaner too 😅

PR #230, opened 2023-12-20
Add dependabot config for version updates

### LINKED ISSUE #229: [SL-1460] Add Dependabot config to dbt-semantic-interfaces
null

<sub>[SL-1460](https://linear.app/dbt-labs/issue/SL-1460/add-dependabot-config-to-dbt-semantic-interfaces)</sub>

tlento:
Instead of manually maintaining version dependencies, let's see
how the robot does.

This should be tuned to produce 1 big update PR per week. That
should minimize the initial noise while we figure out the
appropriate configurations for this tool.

Resolves #229

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,26 @@
+version: 2
+updates:
+  # python dependencies
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


QMalcolm (approved):
Looks good to me, thank you for doing this work!

PR #238, opened 2023-12-21
Begin supporting Pydantic 2 (without dropping support for Pydantic 1)

QMalcolm:
Resolves #134 

### Description

After community and internal discussions we decided the best path forward was to support both Pydantic 1 and Pydantic 2 for the time being as we can't drop Pydantic 1 yet, but we also need to allow for Pydantic 2. We decided the tidiest way to do this would be via shimming. This PR adds the shim, directs pydantic imports to it, loosens the pydantic requirement to allow for pydantic 2, and ensure our testing checks both moving forward.

Also, I want shout out other PRs that worked to address this issue #217 and #227 as well as the people who worked on them (@esciara and @bernardcooke53 respectively). Their work was fundamental to making the changes that this PR contains, and they should be considered contributors on this effort.

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #240, opened 2024-01-03
[BACKPORT] #238 to 0.4.latest (Pydantic 2 Support)

QMalcolm:
This is a backport of #238 to 0.4.latest. The commits included in this PR were cherry-picked from commits added to main from #238. No changes were made, there were no merge conflicts.

### DIFF
--- a/.changes/unreleased/Dependencies-20231221-130418.yaml
+++ b/.changes/unreleased/Dependencies-20231221-130418.yaml
@@ -1,0 +1,6 @@
+kind: Dependencies
+body: Add support for pydantic 2 (and continue support for pydantic 1)
+time: 2023-12-21T13:04:18.792592-08:00
+custom:
+  Author: QMalcolm bernardcooke53 esciara
+  PR: "134"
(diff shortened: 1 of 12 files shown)


MichelleArk (approved):
Looks good to me!

Thinking out loud through any potential risk given this is a backport: 
The change is quite safe and backwards-compatible. The only risk I could imagine is if 0.4 users start installing pydantic 2.0 and the downgrade path via pydantic.v1/pydantic_shim isn't exactly 1:1 with pydantic v1. 

But given[ pydantic.v1 is documented for this very purpose,](https://docs.pydantic.dev/2.0/migration/#install-pydantic-v2) I'd be surprised if that were the case! Additionally, given the additional testing you've added to run against both major pydantic versions, it feels like we've mitigated this risk sufficiently 👍

QMalcolm:
So there is some internal risk. The only thing dbt-core uses pydantic for out of the box is the production of the semantic manifest. I plan to address this risk by releasing a 0.4.3dev0 version of DSI and manually testing that an pydantic 1 and a pydantic 2 core environment produce the same semantic manifest.

For the open source community there should be essentially zero risk. All this will be is a quality of life improvement for them. As it will allow people using pydantic 2 and stuck on dbt-core to upgrade to dbt-core 1.7 and it'll allow people who are on dbt-core 1.7 who want to migrate off of pydantic 1 to pydantic 2 to finally do so.

PR #245, opened 2024-01-08
Update pyproject.toml to show license on PyPI

matialm:
license it's not showing in pypi page

Resolves #

### Checklist

### DIFF
--- a/.changes/unreleased/Docs-20240108-193257.yaml
+++ b/.changes/unreleased/Docs-20240108-193257.yaml
@@ -1,5 +1,5 @@
 kind: Docs
-body: license it's not showing in pypi page
+body: Ensured license information gets displayed in pypi
 time: 2024-01-08T19:32:57.6875672-03:00
 custom:
   Author: matialm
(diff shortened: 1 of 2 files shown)


.changes/unreleased/Docs-20240108-193257.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Docs
+body: license it's not showing in pypi page
```
QMalcolm, line 2: We use changie documents for generating our changelog at release time. With that in mind, we want the message from them to say what was fixed, instead of what was wrong. With that in mind, I think we want to update the `body` here to be something more like `Ensured license information gets displayed in pypi`.

QMalcolm (changes requested):
Thank you for catching this and doing this work! ❤️ We should absolutely get this in to make it clear on PyPi what the license of this package is. One small change request for the changie doc, just to try and keep our changelog uniform. Thanks again!

.changes/unreleased/Docs-20240108-193257.yaml
matialm: perfect I change it, thanks for the feedback

QMalcolm (approved):
Looks great! Thanks again ❤️

PR #262, opened 2024-02-14
Update CI to test Python 3.12 for dbt-semantic-interfaces

tlento:
In order to support Python 3.12 we need to ensure it is fully
tested on the local package, and that we can build and run all
of our tests as expected.

This updates the matrix test accordingly. Note no dependency
adjustment is required, as this package is already open-ended
with respect to maximum supported Python version. For now, we
leave it that way.


marcodamore (approved):

marcodamore:
Lgtm - should we add a line [here](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/pyproject.toml#L19) calling out 3.12?

PR #263, opened 2024-02-15
Update developer CI actions and package metadata to include Python 3.12

tlento:
The initial PR for updating to Python 3.12 was missing a metadata update
for publishing supported version ranges to package distributions.

While updating this I noticed that other actions were explicitly testing against
Python 3.9 due to a dependency on an old composite action hard-coded to
3.9 from way back when the Transform product only supported that Python version.
I've taken the liberty of updating all of these actions to use matrix strategies that
test either the complete set of supported Python versions (unit tests) or the min and
max supported Python versions (development-focused actions).

### DIFF
--- a/.github/actions/setup-python-env/action.yml
+++ b/.github/actions/setup-python-env/action.yml
@@ -1,13 +1,17 @@
 name: Setup Python env
 description: Install Python & Hatch
+inputs:
+  python-version:
+    description: Python version number to install
+    required: true
 
 runs:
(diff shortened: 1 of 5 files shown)


tlento:
The JSON schema action also passes on this PR. See https://github.com/dbt-labs/dbt-semantic-interfaces/actions/runs/7921098902

courtneyholcomb (approved):

PR #264, opened 2024-02-15
Update Jinja2 to >= 3.1.3 to address CVE-2024-22195 

tlento:
We received a [dependabot alert](https://github.com/dbt-labs/dbt-semantic-interfaces/security/dependabot/1)
notifying us of this vulnerability.

Dependabot thought this had been fixed in a different PR, but
that was a false positive. This PR makes the relevant update on top
of an underlying commit to move us off of the ~= dependency syntax.


courtneyholcomb (approved):
Thank you for handling!!

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
   "jsonschema>=4.0,<5",
   "PyYAML>=6.0,<7",
   "more-itertools>=8.0,<11.0",
-  "Jinja2>=3.0,<4",
+  "Jinja2>=3.1.3,<4",
```
QMalcolm, line 29: If we wanted to be looser with it, we could exclude only the offending version using the [version exclusion operator](https://peps.python.org/pep-0440/#version-exclusion) which would make it look something like `Jinja2>=3.0,<4,!=3.1.2`

QMalcolm (approved):
Looks good!

PR #272, opened 2024-03-07
Update dbt-semantic-interfaces to 0.6.0dev0

tlento:
Now that 0.5.0 has been deployed from the 0.5.latest branch, we
should update the version on main to the next base dev target.

We use 0.6.0dev0 because starting at 0 or 1 makes sense but
starting at 2 does not, and I have no idea if the next person
will update the version before deployment.


QMalcolm (approved):
Thank you!

PR #274, opened 2024-03-16
Allow metrics in filters

### LINKED ISSUE #273: [Feature] Support referencing a metric in where filter

### Describe the feature

# **Context**

Some metrics require multiple aggregations or dependencies on other metrics. For example, Streamlit measures “super users” which is calculated as a distinct count of users who have 10+ active days in the last 70 days. Notion measures “activated spaces” which is calculated as a distinct count of spaces where at least two users have two active days within the first 7 days of space creation.  Both of these metrics require dependencies on some aggregation in order to produce the end metric and MetricFlow has no path to supporting this today. This paradigm is sometimes called “metrics as dimensions” by our users where an aggregation is an input to the metric definition.

Our plan to support this enable referencing a metric in a filter.

The general interface update is:

Parameter | Description | Type
-- | -- | --
metric:filter | The filter condition applied to a metric. Previously this required the “Dimension(’dim_name’”) object syntax. You would now be able to reference metric and list of group by parameters i.e “Metrics(metric='', group_by=[])” | Required

 For example in a metric config:
```
metrics:
  - name: The metric name
    type: Categorical or Time or Aggregation
    type_params: specific type params 
    filter: | ##The metric filter. Can now reference a metric and associated dimensions or 
	{{Metric('metric_name', group_by=['list of entities or dimensions])}}
```
In a CLI query:
```
mf query --metrics average_bookings --group-by metric_time --where "{{Metric('bookings', ['user'])}} >2"
```

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

courtneyholcomb:
Resolves #273
Resolves SL-1849

### Description
Allow users to reference metrics in where filters for metrics, measures, and saved queries. This uses syntax like:
`{{ Metric('metric_name', group_by=['entity_name', 'dimension_name']) }} = 10`
This unlocks new types of metrics that users have been asking for. Some examples can be found in the linked issue and in [this design doc](https://www.notion.so/dbtlabs/Metrics-as-Dimensions-55718e9516a7462787ffd6e3e8c1237e?pvs=4).

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
+++ b/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
@@ -3,6 +3,7 @@ from typing import Optional, Sequence
 from dbt_semantic_interfaces.call_parameter_sets import (
     DimensionCallParameterSet,
     EntityCallParameterSet,
+    MetricCallParameterSet,
     ParseWhereFilterException,
     TimeDimensionCallParameterSet,
 )
@@ -14,6 +15,8 @@ from dbt_semantic_interfaces.naming.keywords import (
 from dbt_semantic_interfaces.references import (
     DimensionReference,
     EntityReference,
+    LinkableElementReference,
+    MetricReference,
     TimeDimensionReference,
 )
 from dbt_semantic_interfaces.type_enums import TimeGranularity
@@ -100,4 +103,12 @@ class ParameterSetFactory:
         return EntityCallParameterSet(
             entity_path=additional_entity_path_elements + structured_dundered_name.entity_links,
             entity_reference=EntityReference(element_name=structured_dundered_name.element_name),
+        )
+
+    @staticmethod
+    def create_metric(metric_name: str, group_by: Sequence[str] = ()) -> MetricCallParameterSet:
+        """Gets called by Jinja when rendering {{ Metric(...) }}."""
+        return MetricCallParameterSet(
+            metric_reference=MetricReference(element_name=metric_name),
+            group_by=tuple([LinkableElementReference(element_name=group_by_name) for group_by_name in group_by]),
         )
--- a/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
+++ b/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
@@ -109,7 +109,10 @@ class ParameterSetFactory:
     def create_metric(metric_name: str, group_by: Sequence[str] = ()) -> MetricCallParameterSet:
         """Gets called by Jinja when rendering {{ Metric(...) }}."""
         if not group_by:
-            raise ParseWhereFilterException("`group_by` parameter is required for Metric in where filter.")
+            raise ParseWhereFilterException(
+                "`group_by` parameter is required for Metric in where filter. This is needed to determine 1) the granularity to "
+                "aggregate the metric to and 2) how to join the metric to the rest of the query."
+            )
         return MetricCallParameterSet(
             metric_reference=MetricReference(element_name=metric_name),
             group_by=tuple([LinkableElementReference(element_name=group_by_name) for group_by_name in group_by]),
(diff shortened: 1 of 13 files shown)


dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
```diff
@@ -101,3 +104,11 @@ def create_entity(entity_name: str, entity_path: Sequence[str] = ()) -> EntityCa
             entity_path=additional_entity_path_elements + structured_dundered_name.entity_links,
             entity_reference=EntityReference(element_name=structured_dundered_name.element_name),
         )
+
+    @staticmethod
+    def create_metric(metric_name: str, group_by: Sequence[str] = ()) -> MetricCallParameterSet:
+        """Gets called by Jinja when rendering {{ Metric(...) }}."""
+        return MetricCallParameterSet(
+            metric_reference=MetricReference(element_name=metric_name),
+            group_by=tuple([LinkableElementReference(element_name=group_by_name) for group_by_name in group_by]),
```
courtneyholcomb, line 113: TODO: make sure you can't pass a metric into the metric's group by

dbt_semantic_interfaces/parsing/where_filter/where_filter_entity.py
```diff
@@ -27,6 +33,20 @@ def _implements_protocol(self) -> QueryInterfaceEntity:
...
+
+    QueryInterfaceMetric currently has no methods and the parameter set is created in the factory.
+    """
+
+    @override
+    def _implements_protocol(self) -> QueryInterfaceMetric:
+        return self
+
+    def descending(self, _is_descending: bool) -> QueryInterfaceMetric:  # noqa: D
+        raise InvalidQuerySyntax("descending is invalid in the where parameter and filter spec")
```
tlento, line 47: Ok, I see why this has to be here. We really need to carve out some time to clean up the underlying protocol specs so we don't have these runtime exceptions floating around all over the place.

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
```diff
@@ -108,6 +108,8 @@ def create_entity(entity_name: str, entity_path: Sequence[str] = ()) -> EntityCa
     @staticmethod
     def create_metric(metric_name: str, group_by: Sequence[str] = ()) -> MetricCallParameterSet:
         """Gets called by Jinja when rendering {{ Metric(...) }}."""
+        if not group_by:
+            raise ParseWhereFilterException("`group_by` parameter is required for Metric in where filter.")
```
tlento, line 112: Can we make this a little more descriptive as to why? Maybe something like:

```
"`group_by` parameter is required for filtering on Metrics, since a filter on a Metric does not make sense for queries where there are no sub-groups against which to apply the filter expression."
```

I don't know if normal people would understand that, but maybe it's better?

tests/validations/test_saved_query.py
```diff
@@ -135,3 +135,26 @@ def test_invalid_group_by_format_in_saved_query(  # noqa: D
...
+                    where_filters=[
+                        PydanticWhereFilter(where_sql_template="{{ Metric('bookings', ['listings']) }} > 2")
+                    ],
+                ),
+            ),
+        ),
+    ]
+
+    manifest_validator = SemanticManifestValidator[PydanticSemanticManifest]([SavedQueryRule()])
+    manifest_validator.validate_semantic_manifest(manifest),
```
tlento, line 160: Ah, ok, so this should also run our internal check for missing group-bys and will throw a validation error if someone messes that up, correct?

tlento (approved):
Looks good to me! If you haven't already I think we'll want to do some local testing against the jaffle project to make sure parsing the metric filter is working as expected with the dbt integration for 1.8. It should - as far as I know dbt's parser just passes the where filter expressions into these components - but I've never actually tried it.

dbt_semantic_interfaces/parsing/where_filter/where_filter_entity.py
courtneyholcomb: Definitely agree!

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
courtneyholcomb: Updated!

tests/validations/test_saved_query.py
courtneyholcomb: It didn't quite do that, but now I added some more detail so that it will do that.

PR #284, opened 2024-06-05
Revert "Improve validations for derived metrics"

courtneyholcomb:
Reverts dbt-labs/dbt-semantic-interfaces#279

With the onset of dbt-mantle, we can no longer add new validations that might cause errors for existing manifests, or customers might see jobs abruptly failing. Instead, we need to figure out a new process for rolling out validations - likely adding warnings to start.


WilliamDee (approved):
oof

PR #288, opened 2024-06-10
Add & validate cumulative type params & period agg

### LINKED ISSUE #289: [Feature] Add `period_agg` field for cumulative metrics

### Describe the feature

Add a type_param field called `period_agg` for cumulative metrics. This will be used to support querying various time granularities with cumulative metrics by giving users options for how they want tho aggregate across that time period.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

courtneyholcomb:
Resolves #289
Completes SL-2363
Completes SL-2364

### Description
Move cumulative metric type params to a nested key within `type_params` called `cumulative_type_params`.
Add new `period_agg` field with options `FIRST`, `LAST`, and `AVERAGE`. (Note we chose `FIRST` and `LAST` because that language aligns with the window function syntax we'll use for those - `FIRST_VALUE` and `LAST_VALUE`.) This will be used when aggregating cumulative metrics to non-default time granularities.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/metrics.py
+++ b/dbt_semantic_interfaces/validations/metrics.py
@@ -40,8 +40,9 @@ class CumulativeMetricRule(SemanticManifestValidationRule[SemanticManifestT], Ge
             )
 
             for field in ("window", "grain_to_date"):
+                type_params_field_value = getattr(metric.type_params, field)
                 # Warn that the old type_params structure has been deprecated.
-                if getattr(metric.type_params, field):
+                if type_params_field_value:
                     issues.append(
                         ValidationWarning(
                             context=metric_context,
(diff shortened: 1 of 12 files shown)


dbt_semantic_interfaces/validations/metrics.py
```diff
@@ -33,29 +34,62 @@ def _validate_cumulative_sum_metric_params(metric: Metric) -> List[ValidationIss
...
+                ):
+                    issues.append(
+                        ValidationError(
+                            context=metric_context,
+                            message=(
+                                f"`{field}` set twice in cumulative metric '{metric.name}'. "
+                                "Please remove it from `type_params`."
+                            ),
+                        )
+                    )
```
QMalcolm, line 68: We actually want them to be duplicated I think. In which case we should change this to error if the fields are different.

QMalcolm (changes requested):
This is great! I do think we should probably make some changes. I'm trying to think through both the system as a whole (core, DSI, MetricFlow), DSI by itself, the deprecation/migration path in both of those, and the end user experience.

1. MetricFlow should only have to pull from one place 
  a. Ideally it should be able to pull directly from `metric.type_params.cumulative_type_params.<field>` and not think about `metric_type_params.<deprecated_field>`
  b. DSI can ensure this by either
    1. Including a new transformation (I believe MetricFlow still runs transformations) which sets the new paths if they aren't already set when the deprecated paths are.
    2. providing a `metric` level accessor which does the munging
       * this is problematic in that it creates more "future" work in that we'll want to deprecate the accessor at some point and end up migrating MetricFlow not once but twice.
    3. at post instantition of a metric, populating the cumulative_type_params from the old deprecated fields (only for `cumulative_type_param` fields which are `None`)
2. dbt-core should always populate the `cumulative_type_param` fields if they are available
  a. if the user specifies them directly, perfect
  b. if the user specifies them via the deprecated fields, but not directly, then we use the deprecated fields

Points (1.b) and (2) mean that our desired state for the time being is that there _is_ duplication. Thus both being defined _shouldn't_ be an error, it should be expected. What should be an error is if they are _different_.

Additionally, people who use dbt-core have the ability to `silence` warnings. Which might happen for these warnings as they will likely be noisy. It is basically guaranteed to happen for anybody who currently has a cumulative metric defined. Unfortunately silencing happens via log event type and all DSI validation logs get logged out via the same event type. That means silencing this DSI validation issue will silence all DSI validation issues (not great). This isn't something we should fix in this PR, but it's probably worth SL and Core working together on how to make this better.

courtneyholcomb:
@QMalcolm thank you for the thorough feedback! Just added a couple commits to add the transformation and update the validation if you want to take another look.

QMalcolm (approved):
Thank you for working through the change requests ❤️ This looks great!

Also, thank you for being smarter than me. In my change requests, I said that we should error if the deprecated path values and the new path values differed. The edge case in what I said is that it should be allowable for either to be `None`. How you set up the validation rule handles that 🚀

courtneyholcomb:
Note: We have a release for 0.6.0 coming soon. This should not merge until after that release - otherwise users will get validation warnings for `cumulative_type_params` before those fields are actually supported in MF.

PR #291, opened 2024-06-14
Add `default_granularity` to metric spec

### LINKED ISSUE #290: [Feature] Add `default_grain` to metric spec

### Describe the feature

Now that we are enabling sub-daily granularities, the current behavior of `metric_time` will likely be unexpected to users. If they change any of their time dimensions to use a grain smaller than `DAY`, `metric_time` will default to that grain. To avoid this confusion, add a parameter that allows users to choose which grain they want their metric to default to. This will default to `DAY` if not set to avoid the unexpected behavior change. (If `DAY` is smaller than the defined granularity, will default to the smallest available granularity.)

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

courtneyholcomb:
Resolves #290

### Description
Now that we are enabling sub-daily granularities, the current behavior of `metric_time` will likely be unexpected to users. If they change any of their time dimensions to use a grain smaller than DAY, `metric_time` will default to that grain. To avoid this confusion, add a parameter that allows users to choose which grain they want their metric to default to. This will default to DAY if not set to avoid the unexpected behavior change. (If DAY is smaller than the defined granularity, will default to the smallest available granularity.)

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/metric.py
+++ b/dbt_semantic_interfaces/implementations/metric.py
@@ -211,7 +211,7 @@ class PydanticMetric(HashableBaseModel, ModelWithMetadataParsing, ProtocolHint[M
     metadata: Optional[PydanticMetadata]
     label: Optional[str] = None
     config: Optional[PydanticMetricConfig]
-    default_grain: Optional[TimeGranularity] = None
+    default_granularity: Optional[TimeGranularity] = None
 
     @property
     def input_measures(self) -> Sequence[PydanticMetricInputMeasure]:
--- a/dbt_semantic_interfaces/implementations/metric.py
+++ b/dbt_semantic_interfaces/implementations/metric.py
@@ -16,12 +16,7 @@ from dbt_semantic_interfaces.implementations.filters.where_filter import (
     PydanticWhereFilterIntersection,
 )
 from dbt_semantic_interfaces.implementations.metadata import PydanticMetadata
-from dbt_semantic_interfaces.protocols import (
-    Metric,
-    MetricConfig,
-    MetricInputMeasure,
-    ProtocolHint,
-)
+from dbt_semantic_interfaces.protocols import Metric, MetricConfig, ProtocolHint
 from dbt_semantic_interfaces.references import MeasureReference, MetricReference
 from dbt_semantic_interfaces.type_enums import (
     ConversionCalculationType,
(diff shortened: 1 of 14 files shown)


tlento:
I think this branch needs an update, will hold off until the cumulative metrics pieces are through.

dbt_semantic_interfaces/implementations/metric.py
```diff
@@ -228,3 +238,31 @@ def input_metrics(self) -> Sequence[PydanticMetricInput]:
             return (self.type_params.numerator, self.type_params.denominator)
         else:
             assert_values_exhausted(self.type)
+
+    @staticmethod
+    def all_input_measures_for_metric(
```
WilliamDee, line 243: should this method be in the protocol?

dbt_semantic_interfaces/implementations/metric.py
```diff
@@ -228,3 +238,31 @@ def input_metrics(self) -> Sequence[PydanticMetricInput]:
...
+        """Gets all input measures for the metric, including those defined on input metrics (recursively)."""
+        measures = set()
+        if metric.type is MetricType.SIMPLE or metric.type is MetricType.CUMULATIVE:
+            assert (
+                metric.type_params.measure is not None
+            ), f"Metric {metric.name} should have a measure defined, but it does not."
+            measures.add(metric.type_params.measure)
+        elif metric.type is MetricType.DERIVED or metric.type is MetricType.RATIO:
+            for input_metric in metric.input_metrics:
+                nested_metric = metric_index.get(MetricReference(input_metric.name))
```
WilliamDee, line 255: could just use the `input_metric.as_reference` property

dbt_semantic_interfaces/implementations/metric.py
```diff
@@ -228,3 +238,31 @@ def input_metrics(self) -> Sequence[PydanticMetricInput]:
...
+    def all_input_measures_for_metric(
+        metric: Metric, metric_index: Dict[MetricReference, Metric]
+    ) -> Set[MetricInputMeasure]:
+        """Gets all input measures for the metric, including those defined on input metrics (recursively)."""
+        measures = set()
+        if metric.type is MetricType.SIMPLE or metric.type is MetricType.CUMULATIVE:
+            assert (
+                metric.type_params.measure is not None
+            ), f"Metric {metric.name} should have a measure defined, but it does not."
+            measures.add(metric.type_params.measure)
```
WilliamDee, line 252: Not blocking, but heads up that this could end up with duplicate measures.

Ie.,
if a metric that has 2 measures is the same but has a different input filter, then the `set()` wouldn't count them as duplicates as the `MetricInputMeasure` will have a different `filter` field even tho it's referencing the same measure. 

If you don't want duplicate measures showing up in this, you could return `Set[MeasureReference]` instead.

WilliamDee (approved):

dbt_semantic_interfaces/implementations/metric.py
courtneyholcomb: Good call!!

dbt_semantic_interfaces/implementations/metric.py
courtneyholcomb: I think not. I don't think we want to require implementing this logic any place that implements a metric.

PR #293, opened 2024-06-16
Update time spine configs for sub-daily granularity & custom calendar

### LINKED ISSUE #280: [Feature] Sub-daily granularity

### Describe the feature

Enable sub-daily `TimeGranularity` options.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

courtneyholcomb:
Resolves #280

### Description
As we work to support sub-daily granularities, users will be able to set up multiple time spines at different granularities, and this updates the YAML spec to support what we'll need for that. It also lays the groundwork for adding custom time spines (for the custom calendar feature). The configs have been designed to look like this:
```
semantic_layer_project_configs:
	 time_spines:
		 - name: daily_time_spine
			 model: ref('daily_time_spine') #a model in your dbt project
			 base_column: base_date #exprs don't work here. This is just a column
			 base_granularity: day #takes day, week, month, quarter, year
		 - name: hourly_time_spine
			 model: ref('hourly_time_spine') #a model in your dbt project
			 base_column: base_date #exprs don't work here. This is just a column
			 base_granularity: hour #takes day, week, month, quarter, year
		 - name: custom_time_spine
			 model: ref('custom_time_spine') #a model in your dbt project
			 base_column: base_date #exprs don't work here. This is just a column
			 base_granularity: hour #takes day, week, month, quarter, year
	 		 custom_granularities:
				- name: fortnight # this is the column
					offset_window: 14 days
				- name: fiscal_quarter
					expr: quarter
					offset_window: 90 days
```
This PR implements most of those fields, with the exception of the `custom_granularities` field, for which the design is not finalized (will used for custom calendar).

Note on the changing of field names: Currently, these time spine fields are not configured in YAML (i.e., users don't set them), they are set in core instead. That means we should be able to change the names here as long as we also update them in core and anywhere they are parsed from the manifest and handled downstream, such as in MF or MFS.
Similarly, we want `project_configs` to show up as `semantic_layer_project_configs` for users, but that change can be handled in core.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/semantic_model.py
+++ b/dbt_semantic_interfaces/implementations/semantic_model.py
@@ -18,6 +18,7 @@ from dbt_semantic_interfaces.protocols import (
     SemanticModelConfig,
     SemanticModelDefaults,
 )
+from dbt_semantic_interfaces.protocols.semantic_model import NodeRelation
 from dbt_semantic_interfaces.references import (
     EntityReference,
     LinkableElementReference,
@@ -28,13 +29,17 @@ from dbt_semantic_interfaces.references import (
 from dsi_pydantic_shim import Field, validator
 
 
-class NodeRelation(HashableBaseModel):
+class PydanticNodeRelation(HashableBaseModel, ProtocolHint[NodeRelation]):
     """Path object to where the data should be."""
 
     alias: str
     schema_name: str
     database: Optional[str] = None
     relation_name: str = ""
+
+    @override
+    def _implements_protocol(self) -> NodeRelation:  # noqa: D
+        return self
 
     @validator("relation_name", always=True)
     @classmethod
@@ -57,12 +62,12 @@ class NodeRelation(HashableBaseModel):
         return value
 
     @staticmethod
-    def from_string(sql_str: str) -> NodeRelation:  # noqa: D
+    def from_string(sql_str: str) -> PydanticNodeRelation:  # noqa: D
         sql_str_split = sql_str.split(".")
         if len(sql_str_split) == 2:
-            return NodeRelation(schema_name=sql_str_split[0], alias=sql_str_split[1])
+            return PydanticNodeRelation(schema_name=sql_str_split[0], alias=sql_str_split[1])
         elif len(sql_str_split) == 3:
-            return NodeRelation(database=sql_str_split[0], schema_name=sql_str_split[1], alias=sql_str_split[2])
+            return PydanticNodeRelation(database=sql_str_split[0], schema_name=sql_str_split[1], alias=sql_str_split[2])
         raise RuntimeError(
             f"Invalid input for a SQL table, expected form '<schema>.<table>' or '<db>.<schema>.<table>' "
             f"but got: {sql_str}"
@@ -95,7 +100,7 @@ class PydanticSemanticModel(HashableBaseModel, ModelWithMetadataParsing, Protoco
     name: str
     defaults: Optional[PydanticSemanticModelDefaults]
     description: Optional[str]
-    node_relation: NodeRelation
+    node_relation: PydanticNodeRelation
 
     primary_entity: Optional[str]
     entities: Sequence[PydanticEntity] = []
--- a/.changes/unreleased/Features-20240617-124717.yaml
+++ b/.changes/unreleased/Features-20240617-124717.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Support for configuring multiple time spines at different granularities.
+time: 2024-06-17T12:47:17.06019-07:00
+custom:
+  Author: courtneyholcomb
+  Issue: "280"
--- a/dbt_semantic_interfaces/implementations/project_configuration.py
+++ b/dbt_semantic_interfaces/implementations/project_configuration.py
@@ -14,9 +14,7 @@ from dbt_semantic_interfaces.implementations.semantic_version import (
     UNKNOWN_VERSION_SENTINEL,
     PydanticSemanticVersion,
 )
-from dbt_semantic_interfaces.implementations.time_spine_table_configuration import (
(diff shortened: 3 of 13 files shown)


dbt_semantic_interfaces/implementations/semantic_model.py
```diff
@@ -28,14 +29,18 @@
 from dsi_pydantic_shim import Field, validator
 
 
-class NodeRelation(HashableBaseModel):
+class PydanticNodeRelation(HashableBaseModel, ProtocolHint[NodeRelation]):
```
QMalcolm, line 32: The changes for this class **_are not_** breaking for core

QMalcolm (changes requested):
First, thank you for doing this work ❤️

## Breaking changes _probably_ requiring PR changes
The main issue we need to find a way to address is this change _invalidates all existing semantic manifests_. Any existing semantic manifest won't be parsable by DSI. The "easy" solution is that we just require everyone to run a new `dbt parse`. If that is reasonable, then nothing needs to be done. However, if that isn't reasonable (which I'm leaning towards) then we need a way to be able to parse the old style to populate the new style of time spine specification built into DSI.

## Breaking changes not requiring PR changes
There are a few breaking changes for core that mean we'll need to make sure this goes into the next minor version (`0.7.0` currently, as core already allows for `0.6.*`). Specifically, the breaking changes are:
1. `PydanticTimeSpineTableConfigurations` -> `PydanticTimeSpine` (as well as the shape change)
2. `ProjectConfiguration.time_spine_table_configurations` -> `ProjectConfiguration.time_spines`

Jstein77:
@graciegoheen, I would love to get your feedback on this spec change. We're exposing a new YAML key, `semantic_layer_project_configs`. This is already an internal key we use, but we're now allowing users to set project defaults for their semantic layer. Initially, the only thing they can set up are various time spine models that we need for sub-daily granularity and custom time grains. There are other project defaults that we might make configurable in the future, for example, default timezone information.

I'm picturing users defining their `semantic_layer_project_configs` in the `dbt_project.yml`. An example configuration would look like:

```
name: "jaffle_shop"
version: "1.0.0"
config-version: 2
dbt-cloud: 
    project-id: 263759
model-paths: ["models"]
analysis-paths: ["analyses"]
test-paths: ["tests"]
seed-paths: ["jaffle-data"]
macro-paths: ["macros"]
snapshot-paths: ["snapshots"]
....
semantic_layer_project_configs:
	 time_spines:
		 - name: daily_time_spine
			 model: ref('daily_time_spine') #a model in your dbt project
			 base_column: base_date #exprs don't work here. This is just a column
			 base_granularity: day #takes day, week, month, quarter, year
		 - name: hourly_time_spine
			 model: ref('hourly_time_spine') #a model in your dbt project
			 base_column: base_date #exprs don't work here. This is just a column
			 base_granularity: hour #takes day, week, month, quarter, year
		 - name: custom_time_spine
			 model: ref('custom_time_spine') #a model in your dbt project
			 base_column: base_date #exprs don't work here. This is just a column
			 base_granularity: hour #takes day, week, month, quarter, year
	 		 custom_granularities:
				- name: fortnight # this is the column
					offset_window: 14 days
				- name: fiscal_quarter
					expr: quarter
					offset_window: 90 days
```

graciegoheen:
@Jstein77 When you say "project defaults for their semantic layer" do you mean:
- project defaults, but then you _could_ set overrides for specific metrics, etc. (sorta like [model configs](https://docs.getdbt.com/reference/model-configs)?)
- project defaults, but then you _could_ override them at runtime (sorta like [project variables](https://docs.getdbt.com/docs/build/project-variables)?)
- project defaults that apply to your entire project with no additional override ability (sorta like [behavior change flags](https://docs.getdbt.com/reference/global-configs/legacy-behaviors)?)

Jstein77:
Great question! I really like thinking along the dimensions of resource level overrides, query time level overrides and no overrides. 

For time spines this is more akin to "project defaults that apply to your entire project with no additional override ability".  Once you set up your time spine configurations they are available for you entire project and can't be updated at query time.

graciegoheen:
Got it! What about `dbt-semantic-layer:` then (instead of `semantic_layer_project_configs`)? IMO "project" and "configs" is redundant, and I like having the `dbt` prefix (similar to the `dbt-cloudL` config) so folks don't get confused with the `semantic-models:` config.

Couple of other questions:
- why is `base_column` and `base_granularity` prefixed with `base_`? To distinguish between `custom`?
- `#exprs don't work here. This is just a column` why?
- `base_granularity: hour #takes day, week, month, quarter, year` does `hour` work or no?

Notes:
- we have a few other types of configs that deal with naming a column + granularity (example [here](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#partition-clause))

Jstein77:
Answering your questions in line:

> why is base_column and base_granularity prefixed with base_? To distinguish between custom?

Two reasons:
* The base column should be the finest grain contained in the time spine table. There will likely be multiple time columns in their model, and we wanted to distinguish which one you should use here. We settled on base column and base granularity. 
* To distinguish between custom granularities.

> #exprs don't work here. This is just a column why?

* We don't want to take arbitration SQL for the `base_column`. For time spine tables we prefer you push all the logic to the transformation layer and just expose the final time spine model to us. That's why we decided not to allow exprs here. 

> base_granularity: hour #takes day, week, month, quarter, year does hour work or no?

Yes hour should work here! Sorry we designed this in parallel with sub daily granularity so forgot to update this.

courtneyholcomb:
Closing this PR in favor of creating a new one. New plan to avoid breaking change:
- Add new field `time_spines` that uses new `PydanticTimeSpine` object
- Leave old field & object in place to avoid breaking change
- Consume both in core / MF
- Eventually plan to deprecate old field

PR #294, opened 2024-06-16
Improve validations for derived metrics

### LINKED ISSUE #270: [Bug] Make `expr` required for derived metrics

### Current Behavior

You can define a derived metric without the `expr` property, but if you try to query it you will get an error.

### Expected Behavior

We should have a validation to ensure that the `expr` property is set for derived metrics.

### Steps To Reproduce

Define a derived metric without the `expr` property. Query it.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- dbt-semantic-interfaces:
```

### Additional Context

_No response_

courtneyholcomb:
Resolves #270

### Description
Improve validations for derived metrics.
- If there were no input metrics set, previously users would run into an `AssertionError` that interrupts the semantic manifest validation process. This changes that to be a more user-friendly validation error.
- Add a validation warning to ensure that `expr` is set and that it references all the input metrics.

Note: The first commit here was previously merged and then reverted because adding a `ValidationError` would be a breaking change. This PR updates it to be a `ValidationWarning`.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240524-092249.yaml
+++ b/.changes/unreleased/Fixes-20240524-092249.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Improve validations for derived metrics. Improve error messaging if input metrics
+  are not set. Validate that expr is set and references all input metrics.
+time: 2024-05-24T09:22:49.247603-07:00
+custom:
+  Author: courtneyholcomb
+  Issue: "270"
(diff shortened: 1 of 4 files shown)


WilliamDee (approved):
Did we not do this already? 🤔

courtneyholcomb:
@WilliamDee mentioned in the PR description - ended up reverting that because it was a breaking change to add these as errors! Switched them to warnings here.

PR #297, opened 2024-06-18
Bump dev version

courtneyholcomb:
### Description
Needed to get some new fields into MF (specifically `cumulative_type_params`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.6.1dev0"
+version = "0.6.1.dev0"
 description = 'The shared semantic layer definitions that dbt-core and MetricFlow use'
 readme = "README.md"
 requires-python = ">=3.8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.6.0"
+version = "0.6.1dev0"
 description = 'The shared semantic layer definitions that dbt-core and MetricFlow use'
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.6.0"
+version = "0.6.1dev0"
```
tlento, line 3: This should be 0.6.1.dev0

tlento: Actually, should this be 0.6.1 or should it be 0.7.0.dev0? We're on new feature kind of stuff, right? Whichever works, though.

courtneyholcomb: Following the doc 😅 https://www.notion.so/dbtlabs/DSI-Build-Release-Process-48d94828e8b840abb8529f7501fb2c5b?pvs=4#21b373d41cf040f180648ba211ba374f

courtneyholcomb: I don't think we should bump to 0.7.0 unless it's a breaking change. If we do, it will require another round of bumping version requirements in core which should not be necessary for non-breaking changes.

courtneyholcomb: It looks like the DSI and MF docs differ on the naming practices so that's why it feels like there's thrash here. https://www.notion.so/dbtlabs/Metricflow-Build-Release-Process-81023685d93a45da8211b05dc36b7bf8?pvs=4#9accf24d2929404ba7b9337ad21a33e3

courtneyholcomb: Updated to suit PEP!

tlento (approved):

PR #298, opened 2024-06-20
Adjust application of `cumulative_type_params` to smooth out release process

courtneyholcomb:
### Description
2 adjustments related to `cumulative_type_params`:
- Add `SetCumulativeTypeParamsRule` to the set of transformations that actually get applied in prod & test.
- Disable related validations temporarily. This will allow us to add support for cumulative_type_params in MetricFlow before these validations roll out to customers.

Also upgrades to a new production patch version so that we these changes can be used in `metricflow-semantics` appropriately.

### Checklist

### DIFF
--- a/tests/validations/test_metrics.py
+++ b/tests/validations/test_metrics.py
@@ -43,6 +43,7 @@ from dbt_semantic_interfaces.type_enums import (
     TimeGranularity,
 )
 from dbt_semantic_interfaces.validations.metrics import (
+    CUMULATIVE_TYPE_PARAMS_SUPPORTED,
     ConversionMetricRule,
     CumulativeMetricRule,
     DerivedMetricRule,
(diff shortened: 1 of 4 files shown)


tests/validations/test_metrics.py
```diff
@@ -690,17 +691,25 @@ def test_cumulative_metrics() -> None:  # noqa: D
...
+        expected_substr5 = "Cumulative `type_params.grain_to_date` field has been moved and will soon be deprecated."
+        missing_error_strings = set()
+        for expected_str in [expected_substr1, expected_substr2, expected_substr3, expected_substr4, expected_substr5]:
+            if not any(actual_str.as_readable_str().find(expected_str) != -1 for actual_str in build_issues):
+                missing_error_strings.add(expected_str)
+        assert len(missing_error_strings) == 0, "Failed to match one or more expected issues: "
+        f"{missing_error_strings} in {set([x.as_readable_str() for x in build_issues])}"
+    else:
+        assert len(build_issues) == 1
+        expected_substr1 = "Both window and grain_to_date set for cumulative metric. Please set one or the other."
```
tlento, line 709: Ew. At least this is temporary.... I don't know that we have much choice given the way this test is structured.

If, instead, we could split the old-style test out from the new style we could just do pytest.mark.skip on the new bit and remove the validator from the DEFAULT_RULES list for now, but I don't know how worthwhile that is. It'll make this test cleaner, so it's not a total waste of time.

tlento (approved):
This seems fine for now. The PR is like three changes in one, but they're all so tiny it seems silly to split them up. I decided to edit the PR title instead of thinking about this. Seemed like the best balance.

I leave it up to you if you prefer to restructure that test to clean up the conditional handling or just merge as-is.

PR #304, opened 2024-07-03
Set `default_granularity` properly for nested metrics

courtneyholcomb:
### Description
Previously, this transformation was not setting default granularity properly for metrics with input metrics (i.e., derived & ratio metrics). This is because `Metric.measure_references` only gets the top-level measures, not nested measures from input metrics. This PR fixes that bug by utilizing `PydanticMetric.all_input_measures_for_metric()` instead.
Note that this feature (`default_granularity`) has not been released yet so no customers should be impacted by this bug.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/transformations/default_granularity.py
+++ b/dbt_semantic_interfaces/transformations/default_granularity.py
@@ -1,13 +1,16 @@
-from typing import Set
+from typing import Dict, Set
 
 from typing_extensions import override
 
+from dbt_semantic_interfaces.implementations.metric import PydanticMetric
 from dbt_semantic_interfaces.implementations.semantic_manifest import (
     PydanticSemanticManifest,
 )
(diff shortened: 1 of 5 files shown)


WilliamDee (approved):
Should we make it so that that method shows all measures for derived metrics? I would think it might break something tho

courtneyholcomb:
> Should we make it so that that method shows all measures for derived metrics? I would think it might break something tho

@WilliamDee 🤔 doesn't it do that already?? That's why this fix worked!

WilliamDee:
> > Should we make it so that that method shows all measures for derived metrics? I would think it might break something tho
> 
> @WilliamDee 🤔 doesn't it do that already?? That's why this fix worked!

nvm i mixed it with the other function (`all_input_measures_for_metric`) you created!

PR #305, opened 2024-07-03
Default granularity bug fix: handle time dimensions with the same name

courtneyholcomb:
### Description
Found another bug in SetDefaultGranularityRule. If there are multiple `agg_time_dimensions` with the same name across semantic models, and they have different granularities, we'll only check one. Fixed here!

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/transformations/default_granularity.py
+++ b/dbt_semantic_interfaces/transformations/default_granularity.py
@@ -1,4 +1,5 @@
-from typing import Dict, Set
+import logging
+from typing import Dict, Set, Tuple
 
 from typing_extensions import override
 
@@ -8,6 +9,7 @@ from dbt_semantic_interfaces.implementations.semantic_manifest import (
 )
 from dbt_semantic_interfaces.protocols import ProtocolHint
 from dbt_semantic_interfaces.protocols.metric import Metric
+from dbt_semantic_interfaces.protocols.semantic_model import SemanticModel
 from dbt_semantic_interfaces.references import (
     DimensionReference,
     MetricReference,
@@ -17,6 +19,8 @@ from dbt_semantic_interfaces.transformations.transform_rule import (
     SemanticManifestTransformRule,
 )
 from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
+
+logger = logging.getLogger(__name__)
 
 
 class SetDefaultGranularityRule(ProtocolHint[SemanticManifestTransformRule[PydanticSemanticManifest]]):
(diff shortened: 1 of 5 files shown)


dbt_semantic_interfaces/transformations/default_granularity.py
```diff
@@ -34,7 +38,7 @@ def transform_model(semantic_manifest: PydanticSemanticManifest) -> PydanticSema
                 continue
 
             default_granularity = TimeGranularity.DAY
-            seen_agg_time_dimensions: Set[TimeDimensionReference] = set()
+            seen_agg_time_dimensions: Set[Tuple[SemanticModel, TimeDimensionReference]] = set()
```
WilliamDee, line 41: Im always wary of using larger objects in sets, so if we could change this to like a semanticmodelreference would be nice

WilliamDee (approved):

dbt_semantic_interfaces/transformations/default_granularity.py
courtneyholcomb: Good call

dbt_semantic_interfaces/transformations/default_granularity.py
```diff
@@ -49,11 +53,16 @@ def transform_model(semantic_manifest: PydanticSemanticManifest) -> PydanticSema
                     except AssertionError:
                         # This indicates the agg_time_dimension is misconfigured, which will fail elsewhere.
                         # Do nothing here to avoid disrupting the validation process.
+                        logger.warning(
+                            f"Agg time dimension '{agg_time_dimension_ref.element_name}' not found in model."
+                            "This should raise a validation error elsewhere."
```
plypaul, line 58: Where is it supposed to raise a validation error?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "dbt-semantic-interfaces"
-version = "0.6.3"
+version = "0.6.4"
```
plypaul, line 3: Can you move this to a separate PR?

plypaul (approved):

pyproject.toml
courtneyholcomb: Shoot, sorry got Will's approval and merged before I got this 😅

dbt_semantic_interfaces/transformations/default_granularity.py
courtneyholcomb: https://github.com/dbt-labs/dbt-semantic-interfaces/blob/49a0347597b60fbec8d17771fa561c858f06b046/dbt_semantic_interfaces/validations/agg_time_dimension.py#L47-L48

PR #309, opened 2024-07-08
Remove inconsistent Jinja param errors

courtneyholcomb:
### Description
Remove some errors related to Jinja objects. This will allow more flexible syntax. The main problem driving this change is that we have multiple places that parse the same Jinja syntax, but each has different behavior. This means that if you pass a certain syntax into a where filter, that same syntax won't work in a group by. In order to align the two, we must use the more permissive option, since adding new restrictions would be a breaking change. Changes here:
- Allow passing the grain in the name of a time dimension using dunder syntax. This means it can be passed in the name or via a separate parameter, and we will only error if both are passed and they don't match.
- Don't error if `metric_time` is passed into a `Dimension(...)` object. This error creates inconsistent restrictions, not only between where & group by params, but within this where param. Before this change, it would only error if you passed `metric_time` into `Dimension(...)` _without_ a grain. If you include a grain, no error. Instead of raising this error, just allow it and MF will sort it out.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/call_parameter_sets.py
+++ b/dbt_semantic_interfaces/call_parameter_sets.py
@@ -16,15 +16,16 @@ from dbt_semantic_interfaces.type_enums.date_part import DatePart
 
 @dataclass(frozen=True)
 class DimensionCallParameterSet:
-    """When 'Dimension(...)' is used in the Jinja template of the where filter, the parameters to that call."""
+    """When 'Dimension(...)' is used in a Jinja template, the parameters to that call."""
 
     entity_path: Tuple[EntityReference, ...]
     dimension_reference: DimensionReference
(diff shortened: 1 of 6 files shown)


dbt_semantic_interfaces/call_parameter_sets.py
```diff
@@ -16,15 +16,16 @@
 
 @dataclass(frozen=True)
 class DimensionCallParameterSet:
-    """When 'Dimension(...)' is used in the Jinja template of the where filter, the parameters to that call."""
+    """When 'Dimension(...)' is used in a Jinja template, the parameters to that call."""
```
courtneyholcomb, line 19: Removed specificity to where filter here, since Jinja parsing is used elsewhere now, too.

DevonFulcher (approved):
Thank you!

PR #310, opened 2024-07-11
Rename `Metric.default_granularity` field to `Metric.time_granularity`

courtneyholcomb:
### Description
After some discussion with product and dev experience, we've decided to change the name of this field. It hasn't shipped to core yet so is not a breaking change. The name change is intended to remove the word "default", which should be implied for a YAML field. This is more consistent with other dbt YAML objects, which have a default in YAML that can be overridden at query time. The name `time_granularity` is also more consistent with the field we have on time dimensions.
Everything in this PR is just a find & replace operation.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20240711-105519.yaml
+++ b/.changes/unreleased/Under the Hood-20240711-105519.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Rename Metric.default_granularity field to Metric.time_granularity.
+time: 2024-07-11T10:55:19.887705-07:00
+custom:
+  Author: courtneyholcomb
+  Issue: "290"
--- a/.changes/unreleased/Features-20240614-140515.yaml
+++ b/.changes/unreleased/Features-20240614-140515.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Add default_granularity to metric spec.
+body: Add time_granularity to metric spec.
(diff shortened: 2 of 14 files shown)


.changes/unreleased/Under the Hood-20240711-105519.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
```
DevonFulcher, line 1: Is this a breaking change if someone (other than dbt Core) is depending on DSI?

DevonFulcher (approved):

.changes/unreleased/Under the Hood-20240711-105519.yaml
courtneyholcomb: MF is depending on the change but will be forced to change the references when upgrading DSI, so that's fine. MFS relies on DSI through MF. Anywhere else you're thinking of ?

DevonFulcher: I don't know, does anybody use OSS DSI other than us?

courtneyholcomb: I think it should be fine regardless - the main concern we had was breaking people's manifests. Any other server should be fine to upgrade and migrate accordingly!

PR #317, opened 2024-07-25
Remove name from time spine config

courtneyholcomb:
### Description
Remove `name` field from `TimeSpine` objects. The designs have changed so that there is no longer a name separate from the time spine's model. The model alias will be used instead, which makes this field redundant.
This won't be a breaking change for manifests because none of them have this object yet. (Only an empty list in the `time_spines` field.)

### Checklist


tlento (approved):
> This won't be a breaking change for manifests because none of them have this object yet. (Only an empty list in the time_spines field.)

🤞

PR #322, opened 2024-08-01
Streamline object-builder text parsing

plypaul:
### Description

When parsing the Jinja template in where-filters, there is currently a set of factory and factory-specific classes to implement a specific behavior. I've noticed that there are multiple implementations of those classes (if you add the ones in this repo and dependent repos), and it seemed like there was an opportunity to streamline the structure to handle the update for saved-queries and other changes down the road.

This PR adds a `QueryItemDescription` that encapsulates what the user has specified in the text input, and also adds a `QueryItemTextProcessor` that allows you to:

* Find all the `QueryItemDescription`s in a template.
* Render a template for a given `QueryItemDescription` using a configurable format.
* Configure what builder methods (e.g. `.descending()`) are allowed in which cases.

These features seem like it would support existing use cases, but tagging @DevonFulcher to confirm. 

Also, please view by commit. The first commit includes the interface, so if there are issues there, it would be better to get those resolved before updating the later commits.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/parsing/text_input/rendering_helper.py
+++ b/dbt_semantic_interfaces/parsing/text_input/rendering_helper.py
@@ -1,0 +1,224 @@
+from __future__ import annotations
+
+import typing
+from typing import Callable, FrozenSet, Optional, Sequence
+
+from typing_extensions import override
+
+from dbt_semantic_interfaces.parsing.text_input.ti_description import (
+    ObjectBuilderMethod,
+    QueryItemDescription,
+    QueryItemType,
+)
+from dbt_semantic_interfaces.parsing.text_input.ti_exceptions import (
+    InvalidBuilderMethodException,
+)
+
+if typing.TYPE_CHECKING:
+    from dbt_semantic_interfaces.parsing.text_input.ti_processor import (
+        QueryItemDescriptionProcessor,
+    )
+
+from dbt_semantic_interfaces.parsing.text_input.valid_method import ValidMethodMapping
+
+
+class ObjectBuilderJinjaRenderHelper:
+    """Helps to build the methods that go into the Jinja template `.render()` call.
+
+    e.g.
+
+        SandboxedEnvironment(undefined=StrictUndefined)
+            .from_string(jinja_template)
+            .render(
+                Dimension=render_helper.get_function_for_dimension(),
+                TimeDimension=render_helper.get_function_for_time_dimension(),
+                Entity=render_helper.get_function_for_entity(),
+                Metric=render_helper.get_function_for_metric(),
+            )
+        )
+    """
+
+    def __init__(  # noqa: D107
+        self,
+        description_processor: QueryItemDescriptionProcessor,
+        valid_method_mapping: ValidMethodMapping,
+    ) -> None:
+        self._description_processor = description_processor
+        self._valid_method_mapping = valid_method_mapping
+
+    def get_function_for_dimension(self) -> Callable:
+        """Returns the function that should be passed in to `.render(Dimension=...)`."""
+        description_processor = self._description_processor
+        item_type = QueryItemType.DIMENSION
+        allowed_methods = self._valid_method_mapping[item_type]
+
+        def _create(name: str, entity_path: Sequence[str] = ()) -> _RenderingClassForJinjaTemplate:
+            return _RenderingClassForJinjaTemplate(
+                description_processor=description_processor,
+                allowed_methods=allowed_methods,
+                initial_item_description=QueryItemDescription(
+                    item_type=item_type,
+                    item_name=name,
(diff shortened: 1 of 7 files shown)


dbt_semantic_interfaces/parsing/text_input/rendering_helper.py
```diff
@@ -0,0 +1,224 @@
...
+        args = (time_granularity_name, date_part_name, descending)
+        assert sum(1 for arg in args if arg is not None) == 1, f"Expected exactly 1 argument set, but got {args}"
+
+        builder_method: Optional[ObjectBuilderMethod]
+        if time_granularity_name is not None:
+            builder_method = ObjectBuilderMethod.GRAIN
+        elif date_part_name is not None:
+            builder_method = ObjectBuilderMethod.DATE_PART
+        elif descending is not None:
+            builder_method = ObjectBuilderMethod.DESCENDING
```
DevonFulcher, line 193: Can we somehow remove the necessity for this block? I think we should strive to reduce the amount of boilerplate necessary when adding a builder method.

dbt_semantic_interfaces/parsing/text_input/valid_method.py
```diff
@@ -0,0 +1,31 @@
...
+    DEFAULT_MAPPING_FOR_ORDER_BY: ValidMethodMapping = {
+        QueryItemType.METRIC: frozenset({ObjectBuilderMethod.DESCENDING}),
+        QueryItemType.ENTITY: frozenset({ObjectBuilderMethod.DESCENDING}),
+        QueryItemType.DIMENSION: frozenset(
+            {ObjectBuilderMethod.DESCENDING, ObjectBuilderMethod.GRAIN, ObjectBuilderMethod.DATE_PART}
+        ),
+        QueryItemType.TIME_DIMENSION: frozenset({ObjectBuilderMethod.DESCENDING}),
+    }
+
+    DEFAULT_MAPPING: ValidMethodMapping = {
```
DevonFulcher, line 24: I'm a bit confused by this. Are these the allowed methods for group-by & where-filters?

dbt_semantic_interfaces/parsing/text_input/rendering_helper.py
plypaul: Yeah, this block can be removed. However, I don't think net boiler plate will change - did you have a particular approach in mind?

plypaul: Yeah, this block can be removed. However, I don't think net boiler plate will change - did you have a particular approach in mind?

dbt_semantic_interfaces/parsing/text_input/valid_method.py
plypaul: Yeah, that's what it seems like from the implementation in DSI.

dbt_semantic_interfaces/parsing/text_input/rendering_helper.py
plypaul: Added a commit with that update.

dbt_semantic_interfaces/parsing/text_input/valid_method.py
plypaul: e.g. https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/dbt_semantic_interfaces/parsing/where_filter/where_filter_time_dimension.py

dbt_semantic_interfaces/parsing/text_input/rendering_helper.py
DevonFulcher: The change you made is great thank you! Now if we add another builder method, we don't have to add another `elif`.

dbt_semantic_interfaces/parsing/text_input/valid_method.py
DevonFulcher: Got it thanks. I'm unsure if these mappings are correct then. Does it make sense to have `descending` in group-by & where-filter? Similarly, does it make sense to have `grain` in order-by? That may be worth reconsidering.

DevonFulcher (approved):
Looks great thank you!

dbt_semantic_interfaces/parsing/text_input/valid_method.py
plypaul: > Does it make sense to have descending in group-by & where-filter?

To clarify, are you suggesting that we allow descending in group-by & where-filter? Same thing for the grain.

DevonFulcher: I'm not sure what is supported right now, but I was just thinking logically, does it make sense for users to define `descending` & `grain` in those use cases?

PR #340, opened 2024-08-29
Update custom granularity spec

courtneyholcomb:
### Description
Rename `custom_granularity_columns` -> `custom_granularities` and add `column_name` property. This matches the newly-approved spec that will be added to core.
This would be a breaking change, but since this isn't used anywhere yet, nothing will break.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/protocols/time_spine.py
+++ b/dbt_semantic_interfaces/protocols/time_spine.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import Protocol, Sequence
+from typing import Optional, Protocol, Sequence
 
 from dbt_semantic_interfaces.implementations.node_relation import NodeRelation
 from dbt_semantic_interfaces.type_enums import TimeGranularity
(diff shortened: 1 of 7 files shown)


dbt_semantic_interfaces/protocols/time_spine.py
```diff
@@ -28,7 +28,7 @@ def primary_column(self) -> TimeSpinePrimaryColumn:
 
     @property
     @abstractmethod
-    def custom_granularity_columns(self) -> Sequence[TimeSpineCustomGranularityColumn]:
+    def custom_granularities(self) -> Sequence[TimeSpineCustomGranularityColumn]:
```
tlento, line 31: Don't we need to add column_name to the protocol defined on line 52?

tlento (commented):
Is `column_name` supposed to be part of the protocol as well? I imagine it is so we can construct the query.

dbt_semantic_interfaces/protocols/time_spine.py
courtneyholcomb: Oops you're right!

courtneyholcomb: Updated!

WilliamDee (approved):

PR #346, opened 2024-09-09
Enable custom granularity inputs for filters

### LINKED ISSUE #345: Allow filter parameters to reference custom granularities

tlento:
We are enabling support for custom granularity, which means
users will need to be able to reference them in where filters and
other contexts where their input representations of their manifest
elements will need to be parsed.

Historically, we have restricted time granularities to an enumerated
set of standard supported grains. With the addition of custom
granularities we must now admit the possibility of arbitrary string
names representing valid time granularities.

This change updates the interfaces for parsing time dimensions out
of filter expressions to allow for arbitrary string inputs representing
custom granularity names.

For various reasons - mainly having to do with the filter parser
not having access to the semantic manifest - we will NOT be able
to support custom granularity references via dundered string names. So
something like `metric_time__martian_year` will not work, but
the explicit syntax of `Dimension('metric_time').grain('martian_year')`
will be accepted and processed.

Resolves #345

### DIFF
--- a/dbt_semantic_interfaces/parsing/text_input/ti_description.py
+++ b/dbt_semantic_interfaces/parsing/text_input/ti_description.py
@@ -7,7 +7,6 @@ from typing import Optional, Tuple
 from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
 from dbt_semantic_interfaces.errors import InvalidQuerySyntax
 from dbt_semantic_interfaces.naming.dundered import StructuredDunderedName
-from dbt_semantic_interfaces.type_enums import TimeGranularity
 from dbt_semantic_interfaces.type_enums.date_part import DatePart
 
 
@@ -65,21 +64,13 @@ class ObjectBuilderItemDescription:
                 raise InvalidQuerySyntax("The entity path should not be specified for a metric.")
             if len(structured_item_name.entity_links) > 0:
                 raise InvalidQuerySyntax("The name of the metric should not have entity links.")
-        # Check that dimensions / time dimensions have a valid time granularity / date part.
+        # Check that dimensions / time dimensions have a valid date part.
         elif item_type is QueryItemType.DIMENSION or item_type is QueryItemType.TIME_DIMENSION:
-            if self.time_granularity_name is not None:
-                valid_time_granularity_names = set(time_granularity.value for time_granularity in TimeGranularity)
-                if self.time_granularity_name.lower() not in valid_time_granularity_names:
-                    raise InvalidQuerySyntax(
-                        f"{self.time_granularity_name!r} is not a valid time granularity. Valid values are"
-                        f" {valid_time_granularity_names}"
-                    )
-
             if self.date_part_name is not None:
                 valid_date_part_names = set(date_part.value for date_part in DatePart)
                 if self.date_part_name.lower() not in set(date_part.value for date_part in DatePart):
                     raise InvalidQuerySyntax(
-                        f"{self.time_granularity_name!r} is not a valid time granularity. Valid values are"
+                        f"{self.date_part_name!r} is not a valid date part. Valid values are"
                         f" {valid_date_part_names}"
                     )
 
--- a/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
+++ b/dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
@@ -16,12 +16,19 @@ from dbt_semantic_interfaces.references import (
     MetricReference,
     TimeDimensionReference,
 )
-from dbt_semantic_interfaces.type_enums import TimeGranularity
 from dbt_semantic_interfaces.type_enums.date_part import DatePart
 
 
 class ParameterSetFactory:
-    """Creates parameter sets for use in the Jinja sandbox."""
+    """Creates parameter sets for use in the Jinja sandbox.
+
+    This is ONLY to be used for parsing where filters, and then only for the purposes of extracting
+    some basic information about which elements are being accessed in the filter expression in the
+    small number of contexts where a more complete semantic layer implementation is not available.
+
+    In practice, today, this is used by the dbt core parser, which cannot take on a MetricFlow
+    dependency, in order to provide some DAG annotations around elements referenced in where filters.
+    """
 
     @staticmethod
     def _exception_message_for_incorrect_format(element_name: str) -> str:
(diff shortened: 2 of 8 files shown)


dbt_semantic_interfaces/parsing/text_input/ti_description.py
```diff
@@ -65,21 +64,13 @@ def __post_init__(self) -> None:  # noqa: D105
...
-        # Check that dimensions / time dimensions have a valid time granularity / date part.
+        # Check that dimensions / time dimensions have a valid date part.
         elif item_type is QueryItemType.DIMENSION or item_type is QueryItemType.TIME_DIMENSION:
-            if self.time_granularity_name is not None:
-                valid_time_granularity_names = set(time_granularity.value for time_granularity in TimeGranularity)
-                if self.time_granularity_name.lower() not in valid_time_granularity_names:
-                    raise InvalidQuerySyntax(
-                        f"{self.time_granularity_name!r} is not a valid time granularity. Valid values are"
-                        f" {valid_time_granularity_names}"
-                    )
```
DevonFulcher, line 76: I am not very familiar with custom time granularities. However, it seems like we could still validate that the time granularity used at query time matches a custom one defined by the analytics engineer, right?

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
```diff
@@ -37,26 +44,46 @@ def create_time_dimension(
...
         if len(group_by_item_name.entity_links) != 1 and not is_metric_time_name(group_by_item_name.element_name):
             raise ParseWhereFilterException(
                 ParameterSetFactory._exception_message_for_incorrect_format(time_dimension_name)
             )
-        grain_parsed_from_name = group_by_item_name.time_granularity
-        grain_from_param = TimeGranularity(time_granularity_name) if time_granularity_name else None
+        grain_parsed_from_name = (
+            group_by_item_name.time_granularity.value if group_by_item_name.time_granularity else None
+        )
+        grain_from_param = time_granularity_name
```
DevonFulcher, line 70: Is this line necessary? Why not just use `time_granularity_name` instead of using a different variable with the same value?

DevonFulcher, line 70: Is this line necessary? Why not just use `time_granularity_name` instead of using a different variable with the same value?

DevonFulcher, line 70: Is this line necessary? Why not just use `time_granularity_name` instead of using a different variable with the same value?

dbt_semantic_interfaces/parsing/text_input/ti_description.py
courtneyholcomb: I would love to do this too, is it too complex based on the need to pass around the manifest lookup @tlento?

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
```diff
@@ -16,12 +16,19 @@
...
 )
-from dbt_semantic_interfaces.type_enums import TimeGranularity
 from dbt_semantic_interfaces.type_enums.date_part import DatePart
 
 
 class ParameterSetFactory:
-    """Creates parameter sets for use in the Jinja sandbox."""
+    """Creates parameter sets for use in the Jinja sandbox.
+
+    This is ONLY to be used for parsing where filters, and then only for the purposes of extracting
```
courtneyholcomb, line 25: Is this correct? We also parse the jinja in 1) the JDBC and 2) saved query group by inputs. I believe this is used for at least the saved query group by inputs for some parsing in core, maybe related to node dependencies. The JDBC might use the MF logic, but I'm not positive.

courtneyholcomb (commented):
I might not have been clear about this in discussion, but I think it would be a much better user experience if we actually could pass the granularity name via dundered syntax. We would need to pass the manifest lookup around, which is annoying. But it should be possible since we're enforcing name uniqueness, right?

dbt_semantic_interfaces/parsing/text_input/ti_description.py
tlento: Oh I guess maybe we do allow `Dimension('entity__dimension__granularity')` since we rely on the call parameter sets in most places. This is very confusing, but I'm comfortable saying that only works with standard granularities for now, and we should really get rid of all this jinja stuff in the medium to long term.

DevonFulcher (approved):
LGTM!

dbt_semantic_interfaces/parsing/text_input/ti_description.py
courtneyholcomb: Ok I'm fine with not validating for now and merging as-is, but will put up a task for me to investigate this later and see if I can come up with a non-breaking solution.

courtneyholcomb:
> I might not have been clear about this in discussion, but I think it would be a much better user experience if we actually could pass the granularity name via dundered syntax. We would need to pass the manifest lookup around, which is annoying. But it should be possible since we're enforcing name uniqueness, right?

@tlento - reraising my comment here. I would like for us to be able to pass custom grains via dundered syntax, but is that limited by the same issues raised in the other thread on this PR?
Not blocking for merging this. I can include this note in the investigation task I'm putting up for myself.

tlento:
> I would like for us to be able to pass custom grains via dundered syntax, but is that limited by the same issues raised in the other thread on this PR?

It's the same set of issues, and I think "limited" is the correct term - it's not impossible in all cases, but it's very difficult to do in a reasonable way from within this component inside DSI.

We can support the dundered syntax for query parameters today - even where filters - we'd just need to make sure whatever parses them does a conversion from dundered input to what the DSI parser supports.

The hard part is the manifest - filter parameters stored in the YAML can show up in any metric, and that means we'd need to have the list of custom grains before we try to do any parsing.

> I think it would be a much better user experience if we actually could pass the granularity name via dundered syntax. We would need to pass the manifest lookup around, which is annoying. But it should be possible since we're enforcing name uniqueness, right?

I'd avoid the manifest lookup. All we need is the set of custom granularity names.

I suggest pushing a breaking change to this API and changing call_parameter_sets from a property to a function and passing the set of custom granularity constructs in at that point. This only works if none of our YAML parsers use call_parameter_sets in its property format, which I suspect is the case but I'm not 100% sure. I'd use the equivalent of the "ExpandedTimeGranularity" struct instead of string names, because adding that is potentially useful to us in other ways.

One issue with the breaking change - dbt-core 1.9.x is going to roll to beta soon. I'm not sure if it's worth the risk to try to jam this in right before those cutoffs. Worst case we can do a 2-stage shift, where we keep the property but have it only work for standard grains, add a new function to the protocol, cut everything calling it over to the function, and then remove the property. That defers the "breaking change" and allows us to add the new functionality in cloud even if we miss the core 1.9 cut. Source available users of MetricFlow won't be able to use custom calendar with the dunder syntax until 1.10, but they'll still be able to use the feature as long as they always specify the grain explicitly instead of relying on dunder magic.

PR #350, opened 2024-09-27
Add validation warning when filter is applied to conversion input measure

WilliamDee:
Resolves SL-2862

### Description
Since filtering on a conversion input measure is not fully supported, we will add a warning when this is added to let the user know.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/metrics.py
+++ b/dbt_semantic_interfaces/validations/metrics.py
@@ -11,6 +11,7 @@ from dbt_semantic_interfaces.protocols import (
     ConversionTypeParams,
     Dimension,
     Metric,
+    MetricInputMeasure,
     SemanticManifest,
     SemanticManifestT,
     SemanticModel,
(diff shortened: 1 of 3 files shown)


dbt_semantic_interfaces/validations/metrics.py
```diff
@@ -519,17 +522,31 @@ def _validate_measure(measure_reference: MeasureReference, semantic_model: Seman
                     )
                 )
 
+            if input_measure.filter is not None and not is_base_measure:
+                # Filters for conversion measure input is not fully supported.
```
courtneyholcomb, line 526: nit: are*

dbt_semantic_interfaces/validations/metrics.py
```diff
@@ -519,17 +522,31 @@ def _validate_measure(measure_reference: MeasureReference, semantic_model: Seman
...
+                issues.append(
+                    ValidationWarning(
+                        context=MetricContext(
+                            file_context=FileContext.from_metadata(metadata=metric.metadata),
+                            metric=MetricModelReference(metric_name=metric.name),
+                        ),
+                        message=f"For conversion metrics, Measure input {measure.name} has a filter,"
+                        " filter on a conversion input measure is not fully supported.",
```
courtneyholcomb, line 534: maybe let's say "yet"?

courtneyholcomb (approved):
That was so fast 😂 
Thank you also for the cleanup!!
Note that this will need a version bump (prod version) so we can release it and get it into core. And then a core PR that bumps to that version.

WilliamDee:
> Note that this will need a version bump (prod version) so we can release it and get it into core. And then a core PR that bumps to that version.

@courtneyholcomb Are there any other validation tasks? Would it make sense to wait for some of those to be done then we can batch all those up and release it once?

courtneyholcomb:
> > Note that this will need a version bump (prod version) so we can release it and get it into core. And then a core PR that bumps to that version.
> 
> @courtneyholcomb Are there any other validation tasks? Would it make sense to wait for some of those to be done then we can batch all those up and release it once?

@WilliamDee 
Good call, yeah I think there are! Can you put up a task to make sure we don't forget?

PR #353, opened 2024-10-07
Remove legacy time spine warning message

DevonFulcher:
Resolves # SL-2934

### Description

This PR removes the legacy time spine configuration warning in order to make the warning appear in dbt-core as a [behavior change](https://docs.getdbt.com/reference/global-configs/behavior-changes) instead.

### Checklist

### DIFF
--- a/tests/validations/test_time_spines.py
+++ b/tests/validations/test_time_spines.py
@@ -21,6 +21,9 @@ from dbt_semantic_interfaces.implementations.time_spine import (
     PydanticTimeSpineCustomGranularityColumn,
     PydanticTimeSpinePrimaryColumn,
 )
+from dbt_semantic_interfaces.implementations.time_spine_table_configuration import (
+    PydanticTimeSpineTableConfiguration,
+)
 from dbt_semantic_interfaces.test_utils import semantic_model_with_guaranteed_meta
 from dbt_semantic_interfaces.type_enums import (
     AggregationType,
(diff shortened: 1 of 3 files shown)


tests/validations/test_time_spines.py
```diff
@@ -75,38 +74,6 @@ def test_valid_time_spines() -> None:  # noqa: D
...
-        metrics=[],
-        project_configuration=PydanticProjectConfiguration(
-            time_spine_table_configurations=[
-                PydanticTimeSpineTableConfiguration(location="hurrr", column_name="fun_col", grain=TimeGranularity.DAY)
-            ]
-        ),
-    )
-    issues = validator.validate_semantic_manifest(semantic_manifest)
-    assert not issues.has_blocking_issues
-    assert len(issues.warnings) == 1
```
courtneyholcomb, line 106: Could keep this test but update this to say `assert len(issues.warnings) == 1` and remove the line below!

courtneyholcomb (approved):
Looks good - note that we'll need to release a new production version in order to ship this to core. That process is documented here https://www.notion.so/dbtlabs/DSI-Build-Release-Process-48d94828e8b840abb8529f7501fb2c5b?pvs=4

courtneyholcomb:
Actually, we shouldn't do the production release until after code freeze, since mantle could pick up the new version.

tests/validations/test_time_spines.py
DevonFulcher: I assume you mean `assert len(issues.warnings) == 0`, right?

courtneyholcomb: Yes 😆

PR #355, opened 2024-10-10
Downgrade pydantic in dev env

DevonFulcher:
### Description

This PR makes type-checking more consistent for local development. I have had an annoying issue with the type-checker reporting errors that weren't appearing in CI. Initially, I thought this was a problem with mypy, but actually the problem was that I have been using Pydantic 2, but the types in the codebase are only compatible with Pydantic 1. Specifically, I was getting a lot of errors for properties like `extra_detail` that don't have a default value:

```
class ValidationIssue(ABC, BaseModel):
    """The abstract base ValidationIssue class that the specific ValidationIssue classes are built from."""

    message: str
    context: Optional[ValidationContext] = None
    extra_detail: Optional[str]
```

### Checklist


pyproject.toml
```diff
@@ -66,7 +65,8 @@ dependencies = [
   "isort>=5.12,<6",
   "black>=23.3,<24",
   "ruff==0.0.260",
-  "mypy>=1.3,<2",
+  "mypy==1.6.1",
+  "pydantic>=1.10,<2",             # The types in this codebase with mypy checker are only compatible with pydantic 1
```
courtneyholcomb, line 69: Hmmm I wonder if we should also restrict this in prod? Why only change dev?

DevonFulcher: This is a problem only with mypy and the type annotations. This problem doesn't impact the production release version.

courtneyholcomb (approved):

PR #250, opened 2024-01-20
Expose meta for Metric and SemanticModel

WilliamDee:
Resolves #251 

### Description
This PR exposes the `meta` field for Metric and SemanticModel which enables it to be parsed out in the semantic manifest for use in the SL APIs

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #358, opened 2024-10-23
Add meta to measures, entities, and dimensions

DevonFulcher:
### Description

This PR is modeled off of [this](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/250) prior PR. This PR adds meta to measures, entities, and dimensions.
 
### Checklist


dbt_semantic_interfaces/implementations/elements/dimension.py
```diff
@@ -37,6 +42,16 @@ class PydanticDimensionTypeParams(HashableBaseModel):
     validity_params: Optional[PydanticDimensionValidityParams] = None
 
 
+class PydanticDimensionConfig(HashableBaseModel, ProtocolHint[DimensionConfig]):
```
courtneyholcomb, line 45: If these protocols & implementations are all exactly the same structure, I think we should just use the `MetricConfig` protocol & implementation for all of them. Would be nice to remove the duplicate code if it's not needed. @WilliamDee would you foresee any problems with that?
It would be great if we could rename the class to just something more generic like `SemanticLayerElementConfig` or something if we're going to use it on all these elements. That would technically be a breaking change for DSI, but I just checked in dbt-core and it's not used anywhere. So it should be fine.
We would want to bump DSI to the next minor version if we do that. So if we do that, I can walk you through the steps for that.

courtneyholcomb (commented):
Looks good! Left one main point of feedback.

Would love to also get @WilliamDee 's eyes on this just in case because I'm not super familiar with how `meta` & `config` are supposed to look.

Also - whenever we make a schema change, we get the core team's approval first. I'm not sure if that's necessary here since it's just copying what exists on a metric already, but I would recommend asking Jordan to confirm if we should get their eyes on this before merging.

dbt_semantic_interfaces/implementations/elements/dimension.py
WilliamDee: We could do that, but a foreseeable issue comes when or if the Config ever diverges for different elements. Then in that case it may be abit of a pain in the ass to change.

WilliamDee: actually, seems like core just uses `config` as some arbitrary dict, https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/unparsed.py#L615 so i think it's fine to generalize it in DSI

WilliamDee (commented):
this all looks good to me! Just need to resolve the Courtney's comment of whether we want to consolidate the Config to a single class

dbt_semantic_interfaces/implementations/elements/dimension.py
DevonFulcher: Okay, I made this update for the resources in this PR. I will make a follow-up PR to consolidate the other resources. I will wait on Grace's approval to merge this PR.

dbt_semantic_interfaces/protocols/meta.py
```diff
@@ -0,0 +1,12 @@
+from abc import abstractmethod
+from typing import Any, Dict, Protocol
+
+
+class ConfigMeta(Protocol):  # noqa: D
```
courtneyholcomb, line 5: I think it would be better to stick with our existing naming conventions to avoid confusion. (I got confused reading this 😅 ) So then this protocol would be called `SemanticLayerElementConfig`, and the pydantic implementation becomes `PydanticSemanticLayerElementConfig`.

courtneyholcomb (commented):
Requested one more change!

courtneyholcomb:
Also, do you mind searching dbt-mantle for the `MetricConfig` classes that will be removed to make sure they aren't used there? Both the protocol and the implementation. I already checked core.

courtneyholcomb:
> Also, do you mind searching dbt-mantle for the `MetricConfig` classes that will be removed to make sure they aren't used there? Both the protocol and the implementation. I already checked core.

@DevonFulcher I also didn't realize we had `config` on semantic models too! So can you check both dbt-core and dbt-mantle to make sure all the `SemanticModelConfig` and `MetricConfig` classes aren't used?

dbt_semantic_interfaces/protocols/meta.py
DevonFulcher: Good callout!

DevonFulcher:
>  I also didn't realize we had config on semantic models too! So can you check both dbt-core and dbt-mantle to make sure all the SemanticModelConfig and MetricConfig classes aren't used?

@courtneyholcomb Based on my search, it looks like mantle and core have classes with the same name as `SemanticModelConfig` and `MetricConfig`, but those repos are not importing these classes from dsi. Also, to be clear, this PR shouldn't contain any breaking changes. That is my other PR [here](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/361).

courtneyholcomb:
> > I also didn't realize we had config on semantic models too! So can you check both dbt-core and dbt-mantle to make sure all the SemanticModelConfig and MetricConfig classes aren't used?
> 
> @courtneyholcomb Based on my search, it looks like mantle and core have classes with the same name as `SemanticModelConfig` and `MetricConfig`, but those repos are not importing these classes from dsi. Also, to be clear, this PR shouldn't contain any breaking changes. That is my other PR [here](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/361).

@DevonFulcher Sounds like we're good then!! And yeah, sorry, I didn't put this in the appropriate PR 😆

dbt_semantic_interfaces/implementations/element_config.py
```diff
@@ -0,0 +1,18 @@
...
+
+from pydantic import Field
+from typing_extensions import override
+
+from dbt_semantic_interfaces.implementations.base import HashableBaseModel
+from dbt_semantic_interfaces.protocols.meta import PydanticSemanticLayerElementConfig
+from dbt_semantic_interfaces.protocols.protocol_hint import ProtocolHint
+
+
+class SemanticLayerElementConfig(HashableBaseModel, ProtocolHint[PydanticSemanticLayerElementConfig]):
```
courtneyholcomb, line 11: I think you got the names reversed here! This one should be pydantic, the other one is the protocol

dbt_semantic_interfaces/implementations/element_config.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class PydanticSemanticLayerElementConfig(HashableBaseModel, ProtocolHint[SemanticLayerElementConfig]):
+    """PydanticDimension config."""
+
+    @override
+    def _implements_protocol(self) -> SemanticLayerElementConfig:  # noqa: D
+        return self
+
+    meta: Dict[str, Any] = Field(default_factory=dict)
```
theyostalservice, line 18: Would this work if we add more fields to the config and wanted to treat meta as optional?  (For example, if we wanted to allow someone to add either some metadata or a tag.)

(This is not necessarily a blocking question.)

DevonFulcher: Good point! We can make this optional in the future, though, without a breaking change.

WilliamDee: Is this not already optional? In a sense that, it'll default to an empty dict if you didn't provide it anything

DevonFulcher: Yet it is effectively optional because of the default, although not type `Optional`.

WilliamDee (approved):

PR #359, opened 2024-10-24
Validate granularity names in saved query where filters

theyostalservice:
Resolves https://github.com/dbt-labs/dbt-semantic-interfaces/issues/360

### Description

Validate time spines in saved query where filters
    
Validations for time spines in WHERE filters in Saved Queries.
    
This mimics the where filter time spine validation for  metrics and applies it to saved queries.
    
This also bumps the version.  (I assume this is necessary?)

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20241023-180425.yaml
+++ b/.changes/unreleased/Under the Hood-20241023-180425.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Added validation warnings for invalid time spines in where filters of saved queries.
+time: 2024-10-23T18:04:25.235887-07:00
+custom:
+    Author: theyostalservice
+    Issue: "360"
--- a/.changes/unreleased/Under the Hood-20241023-180425.yaml
+++ b/.changes/unreleased/Under the Hood-20241023-180425.yaml
@@ -1,5 +1,5 @@
 kind: Under the Hood
-body: Added validation warnings for invalid time spines in where filters of saved queries.
+body: Added validation warnings for invalid granularity names in where filters of saved queries.
 time: 2024-10-23T18:04:25.235887-07:00
 custom:
     Author: theyostalservice
--- a/dbt_semantic_interfaces/test_utils.py
+++ b/dbt_semantic_interfaces/test_utils.py
@@ -25,6 +25,9 @@ from dbt_semantic_interfaces.implementations.semantic_model import (
 )
 from dbt_semantic_interfaces.parsing.objects import YamlConfigFile
 from dbt_semantic_interfaces.type_enums import MetricType, TimeGranularity
+from dbt_semantic_interfaces.validations.validator_helpers import (
+    SemanticManifestValidationResults,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -169,3 +172,39 @@ def semantic_model_with_guaranteed_meta(
         dimensions=dimensions,
         metadata=metadata,
     )
+
+
+def check_only_one_error_with_message(  # noqa: D
+    results: SemanticManifestValidationResults, target_message: str
+) -> None:
+    assert len(results.warnings) == 0
+    assert len(results.errors) == 1
+    assert len(results.future_errors) == 0
+
+    found_match = results.errors[0].message.find(target_message) != -1
+    # Adding this dict to the assert so that when it does not match, pytest prints the expected and actual values.
+    assert {
+        "expected": target_message,
+        "actual": results.errors[0].message,
+    } and found_match
+
+
+def check_only_one_warning_with_message(  # noqa: D
+    results: SemanticManifestValidationResults, target_message: str
+) -> None:
+    assert len(results.errors) == 0
+    assert len(results.warnings) == 1
+    assert len(results.future_errors) == 0
+
+    found_match = results.warnings[0].message.find(target_message) != -1
+    # Adding this dict to the assert so that when it does not match, pytest prints the expected and actual values.
+    assert {
+        "expected": target_message,
+        "actual": results.warnings[0].message,
+    } and found_match
+
+
+def check_no_errors_or_warnings(results: SemanticManifestValidationResults) -> None:  # noqa: D
+    assert len(results.errors) == 0
+    assert len(results.warnings) == 0
+    assert len(results.future_errors) == 0
--- a/dbt_semantic_interfaces/test_utils.py
+++ b/dbt_semantic_interfaces/test_utils.py
@@ -27,6 +27,7 @@ from dbt_semantic_interfaces.parsing.objects import YamlConfigFile
 from dbt_semantic_interfaces.type_enums import MetricType, TimeGranularity
 from dbt_semantic_interfaces.validations.validator_helpers import (
     SemanticManifestValidationResults,
+    ValidationIssue,
 )
 
 logger = logging.getLogger(__name__)
@@ -174,37 +175,62 @@ def semantic_model_with_guaranteed_meta(
     )
 
 
+def _assert_expected_validation_message(  # noqa: D
+    issues: Sequence[ValidationIssue],
+    message_fragment: str,
+) -> None:
+    found_match = any([issue.message.find(message_fragment) != -1 for issue in issues])
+    # Adding this dict to the assert so that when it does not match, pytest prints the expected and actual values.
+    assert {
+        "expected": message_fragment,
+        "actual_messages": [issue.message for issue in issues],
+    } and found_match
+
+
+def check_expected_issues(  # noqa: D
+    results: SemanticManifestValidationResults,
+    num_expected_errors: int = 0,
+    num_expected_warnings: int = 0,
+    expected_error_msgs: Sequence[str] = [],
+    expected_warning_msgs: Sequence[str] = [],
+) -> None:
+    """Validates the number, type, and content of ValidationIssues.
+
+    Currently assumes zero future_errors as there are no future_errors
+    implemented, but this function can be expanded to cover those if needed.
+    """
+    assert len(results.warnings) == num_expected_warnings
+    assert len(results.errors) == num_expected_errors
+    assert len(results.future_errors) == 0, "validation function expects zero future_errors to be implemented."
+
+    for expected_error_msg in expected_error_msgs:
+        _assert_expected_validation_message(issues=results.errors, message_fragment=expected_error_msg)
+    for expected_warning_msg in expected_warning_msgs:
+        _assert_expected_validation_message(issues=results.warnings, message_fragment=expected_warning_msg)
+
+
 def check_only_one_error_with_message(  # noqa: D
     results: SemanticManifestValidationResults, target_message: str
 ) -> None:
-    assert len(results.warnings) == 0
-    assert len(results.errors) == 1
-    assert len(results.future_errors) == 0
-
-    found_match = results.errors[0].message.find(target_message) != -1
-    # Adding this dict to the assert so that when it does not match, pytest prints the expected and actual values.
-    assert {
-        "expected": target_message,
-        "actual": results.errors[0].message,
-    } and found_match
+    check_expected_issues(
+        results=results,
+        num_expected_errors=1,
+        expected_error_msgs=[target_message],
+    )
 
 
 def check_only_one_warning_with_message(  # noqa: D
     results: SemanticManifestValidationResults, target_message: str
 ) -> None:
-    assert len(results.errors) == 0
-    assert len(results.warnings) == 1
-    assert len(results.future_errors) == 0
-
-    found_match = results.warnings[0].message.find(target_message) != -1
-    # Adding this dict to the assert so that when it does not match, pytest prints the expected and actual values.
-    assert {
-        "expected": target_message,
-        "actual": results.warnings[0].message,
-    } and found_match
+    check_expected_issues(
+        results=results,
+        num_expected_warnings=1,
+        expected_warning_msgs=[target_message],
+    )
 
 
 def check_no_errors_or_warnings(results: SemanticManifestValidationResults) -> None:  # noqa: D
-    assert len(results.errors) == 0
-    assert len(results.warnings) == 0
-    assert len(results.future_errors) == 0
+    # no num arguments required since all defaults are zero
+    check_expected_issues(
+        results=results,
+    )
--- a/dbt_semantic_interfaces/validations/saved_query.py
+++ b/dbt_semantic_interfaces/validations/saved_query.py
@@ -146,12 +146,17 @@ class SavedQueryRule(SemanticManifestValidationRule[SemanticManifestT], Generic[
     def _check_where_timespine(
         saved_query: SavedQuery, custom_granularity_names: list[str]
     ) -> Sequence[ValidationIssue]:
+        where_param = saved_query.query_params.where
+        if where_param is None:
+            return []
+
         issues: List[ValidationIssue] = []
 
         valid_granularity_names = [
             standard_granularity.name for standard_granularity in TimeGranularity
         ] + custom_granularity_names
-        for where_filter in saved_query.query_params.where.where_filters:
+
+        for where_filter in where_param.where_filters:
             for time_dim_call_parameter_set in where_filter.call_parameter_sets.time_dimension_call_parameter_sets:
                 if not time_dim_call_parameter_set.time_granularity_name:
                     continue
--- a/dbt_semantic_interfaces/validations/saved_query.py
+++ b/dbt_semantic_interfaces/validations/saved_query.py
@@ -23,7 +23,6 @@ from dbt_semantic_interfaces.parsing.where_filter.where_filter_parser import (
 )
 from dbt_semantic_interfaces.protocols import SemanticManifestT
 from dbt_semantic_interfaces.protocols.saved_query import SavedQuery
-from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
 from dbt_semantic_interfaces.validations.validator_helpers import (
     FileContext,
     SavedQueryContext,
@@ -31,7 +30,6 @@ from dbt_semantic_interfaces.validations.validator_helpers import (
     SemanticManifestValidationRule,
     ValidationError,
     ValidationIssue,
-    ValidationWarning,
     generate_exception_issue,
     validate_safely,
 )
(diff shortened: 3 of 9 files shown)


theyostalservice:
* **#359** <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-semantic-interfaces/359?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @theyostalservice and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

dbt_semantic_interfaces/validations/saved_query.py
```diff
@@ -136,9 +138,44 @@ def _check_where(saved_query: SavedQuery) -> Sequence[ValidationIssue]:
                         },
                     )
                 )
+            else:
+                issues += SavedQueryRule._check_where_timespine(saved_query, custom_granularity_names)
 
         return issues
 
+    def _check_where_timespine(
```
theyostalservice, line 146: TODO: Add `@staticmethod` decorator

theyostalservice, line 146: For reviewers: This should look like the code in [metrics.py](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/98b5af50ecf4c8d9730b3b24476cbec63e0aaad3/dbt_semantic_interfaces/validations/metrics.py#L251), adjusted a bit for the calling context

tests/validations/test_saved_query.py
```diff
@@ -38,6 +38,21 @@ def check_only_one_error_with_message(  # noqa: D
     } and found_match
 
 
+def check_only_one_warning_with_message(  # noqa: D
```
theyostalservice, line 41: Based on the `check_onlye_one_error_with_message` logic above.  I thought it would make more sense to factor this out and make it reusable than to inline it.

tests/validations/test_saved_query.py
```diff
@@ -87,6 +102,34 @@ def test_invalid_where_in_saved_query(  # noqa: D
     )
 
 
+def test_where_filter_validations_invalid_granularity(  # noqa: D
```
theyostalservice, line 105: based on the [similar test in the metrics validations tests](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/98b5af50ecf4c8d9730b3b24476cbec63e0aaad3/tests/validations/test_metrics.py#L460).

.changes/unreleased/Under the Hood-20241023-180425.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
+body: Added validation warnings for invalid time spines in where filters of saved queries.
```
courtneyholcomb, line 2: I would amend this slightly (and the PR title too) - this is validating `granularities`, not `time spines`. The granularity they put in might reference a standard granularity (not associated with a time spine) or a custom granularity (which is associated with a time spine, though the time spine is not what's being validated here).

dbt_semantic_interfaces/validations/saved_query.py
```diff
@@ -136,9 +138,44 @@ def _check_where(saved_query: SavedQuery) -> Sequence[ValidationIssue]:
...
+                    continue
+                if time_dim_call_parameter_set.time_granularity_name not in valid_granularity_names:
+                    issues.append(
+                        ValidationWarning(
+                            context=SavedQueryContext(
+                                file_context=FileContext.from_metadata(metadata=saved_query.metadata),
+                                element_type=SavedQueryElementType.WHERE,
+                                element_value=where_filter.where_sql_template,
+                            ),
+                            # message=f"Filter for metric `{context.metric.metric_name}` is not valid. "
```
courtneyholcomb, line 171: Did you mean to leave this in?

dbt_semantic_interfaces/validations/saved_query.py
courtneyholcomb, line 146: Same here - can we rename to granularity instead of time spine?

courtneyholcomb: Related to this - can we dedupe this logic? There is a ton reused between here and `metrics.py`. I'm thinking we should move `WhereFiltersAreParseable` to its own file and move most of this logic to that class, maybe in a method called `_validate_saved_query()`. I think you should be able to use `_validate_time_granularity_names()` for both metrics and saved queries and that should contain all the duplicate logic, you just might need to update the context param to be more generic (if possible).

courtneyholcomb: Also, if you do that, then it might make sense to take the manifest as a param for `WhereFiltersAreParseable` on `__init__`, and you can parse out the custom granularity names and store them there. That would also mean anything that uses the custom granularity names won't be a `staticmethod`, but at least the logic that parses out custom granularity names would be encapsulated in a place that makes sense.

tests/validations/test_saved_query.py
```diff
@@ -87,6 +102,34 @@ def test_invalid_where_in_saved_query(  # noqa: D
...
+    manifest.saved_queries = [
+        PydanticSavedQuery(
+            name="Example Saved Query",
+            description="Example description.",
+            query_params=PydanticSavedQueryQueryParams(
+                metrics=["bookings"],
+                group_by=["Dimension('booking__is_instant')"],
+                where=PydanticWhereFilterIntersection(
+                    where_filters=[
+                        PydanticWhereFilter(where_sql_template="{{ TimeDimension('metric_time', 'cool') }}"),
```
courtneyholcomb, line 119: Can we add more filters that are valid to make sure we don't warn incorrectly for those?
Also, I triggered a bug when I added the metric filter validations because I accidentally made these case sensitive for custom granularity names (that's been fixed since then). Can we add a test filter with an uppercase standard granularity like `DAY` to make sure the same thing doesn't happen here?

dbt_semantic_interfaces/validations/saved_query.py
theyostalservice: I'll see what I can do. :-)

theyostalservice: I'm going to resolve this thread, but please look at the commit message for the most recent commit (I'll probably copy-paste it as a comment on the PR, too, just to be safe) because I structured things a little differently.

tests/validations/test_where_filters_are_parseable.py
```diff
@@ -0,0 +1,309 @@
...
+    assert len(issues.warnings) == 1
+    assert "`cool` is not a valid granularity name" in issues.warnings[0].message
+
+
+# ------------------------------------------------------------------------------
+# Saved Query validations
+# ------------------------------------------------------------------------------
+
+
+def test_saved_query_with_happy_filter(  # noqa: D
```
theyostalservice, line 185: @courtneyholcomb -
Addresses:
> Can we add more filters that are valid to make sure we don't warn incorrectly for those?

tests/validations/test_where_filters_are_parseable.py
```diff
@@ -0,0 +1,309 @@
...
+    manifest.saved_queries = [
+        PydanticSavedQuery(
+            name="Example Saved Query",
+            description="Example description.",
+            query_params=PydanticSavedQueryQueryParams(
+                metrics=["bookings"],
+                group_by=["Dimension('booking__is_instant')"],
+                where=PydanticWhereFilterIntersection(
+                    where_filters=[
+                        PydanticWhereFilter(where_sql_template="{{ TimeDimension('metric_time', 'DAY') }}"),
```
theyostalservice, line 224: @courtneyholcomb 
Addresses/resolves:
> Also, I triggered a bug when I added the metric filter validations because I accidentally made these case sensitive for custom granularity names (that's been fixed since then). Can we add a test filter with an uppercase standard granularity like DAY to make sure the same thing doesn't happen here?

theyostalservice:
For visibility - I wanted to make sure the explanation of where I expanded on the feedback from @courtneyholcomb was visible and explain the reasoning, so I'm copying the commit message from the latest update to a comment:

### Centralize Where filter validation + comments
This is not *exactly* what was described in Courtney's comments, but I
think there was a little roughness in that plan once I began implementing
(but I'm happy to change course if reviewers disagree!).

This commit promotes the `WhereFiltersAreParseable` to be
`WhereFiltersAreParseableRule` (a free-standing 'rule' in a separate
file).  It was weird to be passing things in from other classes but
somehow centralizing the manifest, so instead, I just moved ALL of
the relevant checks here.

This moves the tests for where filters to a new file specifically
for this rule (again, I'm open to the idea that this would be better
just divided amongst the existing tests, but they share so much
conceptually that it seems nice to group them together and to have
a test that is pointed 1:1 at a rule where possible).

Finally, I also moved some of the test support functions
(`check_only_one_error_with_message`, for example) to
`dbt_semantic_interfaces.test_utils` because they seem useful and
reusable.

tests/validations/test_where_filters_are_parseable.py
```diff
@@ -0,0 +1,309 @@
...
+    SemanticManifestValidationException,
+)
+from dbt_semantic_interfaces.validations.where_filters import (
+    WhereFiltersAreParseableRule,
+)
+
+logger = logging.getLogger(__name__)
+
+
+# ------------------------------------------------------------------------------
```
theyostalservice, line 36: this section is moved from existing tests.  I renamed a few, and the calls to `deepcopy` became `copy.deepcopy`, but they are not functionally changed.

dbt_semantic_interfaces/test_utils.py
```diff
@@ -169,3 +172,39 @@ def semantic_model_with_guaranteed_meta(
...
+
+def check_only_one_warning_with_message(  # noqa: D
+    results: SemanticManifestValidationResults, target_message: str
+) -> None:
+    assert len(results.errors) == 0
+    assert len(results.warnings) == 1
+    assert len(results.future_errors) == 0
+
+    found_match = results.warnings[0].message.find(target_message) != -1
+    # Adding this dict to the assert so that when it does not match, pytest prints the expected and actual values.
```
theyostalservice, line 200: TODO: figure out if I need to add this to `check_no_errors_or_warnings` and maybe look for a way to avoid repeating the same code 2-3 times across these functions

dbt_semantic_interfaces/validations/where_filters.py
```diff
@@ -0,0 +1,268 @@
...
+    validate_safely,
+)
+
+
+class WhereFiltersAreParseableRule(SemanticManifestValidationRule[SemanticManifestT], Generic[SemanticManifestT]):
+    """Validates that all Metric WhereFilters are parseable."""
+
+    @staticmethod
+    def _validate_time_granularity_names_impl(
+        location_label_for_errors: str,
```
courtneyholcomb, line 28: This param seems kinda hacky - can we instead include 2 params? Something like this:
1. `element_name: str` (the name of the metric/saved query)
2. `object_type: SemanticManifestNodeType` - this would be either `SAVED_QUERY` or `METRIC`
Then you can build the location label as something like `f"{object_type.value} '{element_name}'"`

dbt_semantic_interfaces/validations/where_filters.py
```diff
@@ -0,0 +1,268 @@
...
+            filter_call_parameter_sets_with_context=[
+                (
+                    SavedQueryContext(
+                        file_context=FileContext.from_metadata(metadata=saved_query.metadata),
+                        element_type=SavedQueryElementType.WHERE,
+                        element_value=where_filter.where_sql_template,
+                    ),
+                    where_filter.call_parameter_sets,
+                )
+                for where_filter in where_param.where_filters
```
courtneyholcomb, line 72: Here, we're looping through the where filters once, then we end up looping through them again in `_validate_time_granularity_names_impl()`. This parameter also seems a bit fragile / easy to mess up.
Instead, can we just validate one `WhereFilter` in `_validate_time_granularity_names_impl()`? Then we would only one loop, at the call site. And that would allow us to take 2 clearly-typed params in the function instead of building `filter_call_parameter_sets_with_context`.

dbt_semantic_interfaces/validations/where_filters.py
```diff
@@ -0,0 +1,268 @@
...
+                            custom_granularity_names=custom_granularity_names,
+                        )
+        return issues
+
+    @staticmethod
+    @validate_safely(whats_being_done="running manifest validation ensuring all metric where filters are parseable")
+    def validate_manifest(semantic_manifest: SemanticManifestT) -> Sequence[ValidationIssue]:  # noqa: D
+        issues: List[ValidationIssue] = []
+        custom_granularity_names = [
+            granularity.name
```
courtneyholcomb, line 257: This is a nit because it's not likely significant to performance, but we could include the standard granularity names here as well so that we don't have to loop through them each time we call `_validate_time_granularity_names_impl()`.

courtneyholcomb: Or even better would be to add a cached property using like `WhereFiltersAreParseableRule.valid_granularity_names()`, using `@lru_cache`, to encapsulate this in a nicer way.

tests/validations/test_metrics.py
```diff
@@ -345,145 +342,6 @@ def test_derived_metric() -> None:  # noqa: D
     check_error_in_issues(error_substrings=expected_substrings, issues=build_issues)
 
 
-def test_where_filter_validations_happy(  # noqa: D
```
courtneyholcomb, line 348: Did you change these tests or just move them? With all the logic moved around in this PR, it makes it hard to tell what logic was changed vs. just moved, so I just want to make sure the old tests are still passing!

tests/validations/test_where_filters_are_parseable.py
courtneyholcomb: Ok sweet - this addresses my last comment!

dbt_semantic_interfaces/validations/where_filters.py
```diff
@@ -0,0 +1,268 @@
...
+    generate_exception_issue,
+    validate_safely,
+)
+
+
+class WhereFiltersAreParseableRule(SemanticManifestValidationRule[SemanticManifestT], Generic[SemanticManifestT]):
+    """Validates that all Metric WhereFilters are parseable."""
+
+    @staticmethod
+    def _validate_time_granularity_names_impl(
```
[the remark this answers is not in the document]
theyostalservice: implementation
In previous places, it was a common way to mark an internal function that does "the work" vs functions with more focused purposes that would be called just to organize parameters to pass in, for example.

I'm happy to rename this if you like! :-)

dbt_semantic_interfaces/validations/where_filters.py
theyostalservice: > Instead, can we just validate one WhereFilter in _validate_time_granularity_names_impl()?

I *can*.  The fact that I didn't is probably an artifact of originally avoiding changes to the metrics validation logic since it doesn't use `WhereFilter`s directly.

Buuut I am also not sure it's the right path?  metrics validations use `metric.filter.filter_expression_parameter_sets`.  `metrics.filter` is a WhereFilterIntersection, not a WhereFilter, and there's some parsing error logic baked into its `filter_expression_parameter_sets`.  I can change it out pretty easily, but I lack the context to know if that change is likely to upset folks.

------------

However, maybe a compromise that accomplishes the same thing would be to just validate on a single set of `FilterCallParameterSets` per call to `_validate_time_granularity_names_impl`?  I think that's can be done safely with no meaningful change to how this behaves anywhere.

tests/validations/test_metrics.py
theyostalservice: (resolved below :-) )

dbt_semantic_interfaces/validations/where_filters.py
theyostalservice: (I've messaged you to clarify a few things before I respond here :-) )

theyostalservice: My instinct here is that
1. I do not mind precomputing that here at all
2. I don't think using the LRU is a good idea.  In this case, it's just about code clarity - using a "Least Recently Used Cache" communicates that the contents are incomplete, that we'll be dropping some contents as we add others, etc etc.  That's absolutely not the case here - this is intended to be an exhaustive set, static and useable throughout this entire completely linear run.

courtneyholcomb: Do you have a different idea in mind? Or just prefer to keep it as is?

dbt_semantic_interfaces/validations/semantic_manifest_validator.py
```diff
@@ -86,7 +88,7 @@ class SemanticManifestValidator(Generic[SemanticManifestT]):
         SemanticModelDefaultsRule[SemanticManifestT](),
         PrimaryEntityRule[SemanticManifestT](),
         PrimaryEntityDimensionPairs[SemanticManifestT](),
-        WhereFiltersAreParseable[SemanticManifestT](),
+        WhereFiltersAreParseableRule[SemanticManifestT](),
```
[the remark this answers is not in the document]
theyostalservice: Yuck. :P

> check if it's used in any of the core team's codebases before releasing
Ok, given this:
https://github.com/search?q=org%3Adbt-labs%20WhereFiltersAreParseable&type=code
Is it safe to assume it's not used anywhere else?

If I can just release this batched up with @DevonFulcher , that doesn't sound like a bad trade.  If I can't, then... maybe I don't want to make things complicated over a simple style-unifying name change lol

dbt_semantic_interfaces/validations/where_filters.py
theyostalservice: Oh, sorry, I thought your first suggestion was to add the standard granularities to the computation right here and just pass it through.  For the purpose of simplicity, that would be my preference.

If we really want to stick this in a variable, I'd prefer to make this an instance-based class instead of using static methods and then instantiate WhereFiltersAreParseableRule with a set of valid_granularities, but that's going to create even more noise in this PR.

dbt_semantic_interfaces/validations/where_filters.py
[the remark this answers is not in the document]
courtneyholcomb: Ooo makes sense! I'm not sure people here will get what it means though 😅 at least I didn't! We don't have a great convention for this here except to use the leading underscore.

dbt_semantic_interfaces/validations/where_filters.py
courtneyholcomb: Ooo let me look at the metric filters, I didn't realize they had a different type!

dbt_semantic_interfaces/validations/where_filters.py
courtneyholcomb: Yes, I think that was actually my suggestion in my first round of feedback (to make this instance-based)! This is more of a nit though and I'm fine with just adding the standard granularities here and passing them all through.

dbt_semantic_interfaces/validations/where_filters.py
```diff
@@ -0,0 +1,268 @@
...
+        )
+
+    @staticmethod
+    def _validate_time_granularity_names_for_metric(
+        context: MetricContext,
+        filter_expression_parameter_sets: Sequence[Tuple[str, FilterCallParameterSets]],
+        custom_granularity_names: List[str],
+    ) -> Sequence[ValidationIssue]:
+        return WhereFiltersAreParseableRule._validate_time_granularity_names_impl(
+            location_label_for_errors="metric `{context.metric.metric_name}`",
```
courtneyholcomb, line 84: also should be an f-string?

dbt_semantic_interfaces/validations/semantic_manifest_validator.py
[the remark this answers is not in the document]
theyostalservice: OK after discussion, I'm going to revert the rename.

dbt_semantic_interfaces/validations/where_filters.py
theyostalservice: From discussion: avoid "implementation" as that word is overloaded here.  rename it in next version.

courtneyholcomb (approved):
Looks great thank you!!

PR #363, opened 2024-11-05
Support custom grain in DSI callsites

WilliamDee:
### Description
Given that custom granularity is available, we need to start supporting it when parsing through names and filters.

### Non-breaking changes
- Removed `WhereFilterTimeDimensionFactory` as it's not used anywhere (MetricFlow has it's own in [metricflow-semantics](https://github.com/dbt-labs/metricflow/blob/main/metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_time_dimension.py#L56))
- Added tests to test parsing of group by's using custom grain and filtering with custom grain

### Breaking changes
- Removed `DunderedNameFormatter` in favour of `StructuredDunderedName` as it's duplicate logic
- Added `custom_granularity_names` to `StructuredDunderedName.parse_name` as a parameter to parse out any custom grain
  - Updated callsites for this
- Changed `PydanticWhereFilter.call_parameter_sets` and `PydanticWhereFilterIntersection.filter_expression_parameter_sets` from a property to a method that takes in the valid custom grain names
   - Updated callsites for this 

### Checklist

Resolves SL-2989


dbt_semantic_interfaces/naming/dundered.py
```diff
@@ -91,52 +91,3 @@ def entity_prefix(self) -> Optional[str]:
             return DUNDER.join(tuple(entity_reference.element_name for entity_reference in self.entity_links))
 
         return None
-
-
-class DunderedNameFormatter:
```
courtneyholcomb, line 96: I'm assuming you checked that this isn't used in any dependent repos?

courtneyholcomb: Same for all the other deleted objects!

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
```diff
@@ -49,6 +49,7 @@ def create_time_dimension(
         time_granularity_name: Optional[str] = None,
         entity_path: Sequence[str] = (),
         date_part_name: Optional[str] = None,
+        custom_granularity_names: Sequence[str] = (),
```
courtneyholcomb, line 52: Should this not be optional?

courtneyholcomb (approved):
Left a couple comments but otherwise looks good!

Since this is a breaking change, I would wait to merge it until you're ready to bump the version (otherwise can risk someone else doing a patch version bump that includes your change). This should be a minor version bump.
@DevonFulcher also has a breaking change PR up. Maybe you guys can coordinate releasing & updating core. Make sure you guys sync with me before you update core because there are some extra steps you'll need to take.

dbt_semantic_interfaces/naming/dundered.py
WilliamDee: yup!

WilliamDee: so `DunderedNameFormatter` is used in MF, but I thought it should be fine since this is release a minor version upgrade, which shouldn't be picked up by MF right away right?

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
WilliamDee: felt better for this to be default to a sequence, so we don't need to handle this at the bottom where we do
```python
StructuredDunderedName.parse_name(
            name=time_dimension_name, custom_granularity_names=custom_granularity_names or ()
        )
```

dbt_semantic_interfaces/naming/dundered.py
courtneyholcomb: Sounds good! I would double check that `metricflow-semantics` won't pick it up (it's not always pinned to a patch version)

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
courtneyholcomb: Yeah I guess I just want to make sure people don't skip it without realizing that it's necessary. Like sometimes I like to leave params like this as required so that people are forced to actively pass in the empty value and hopefully in the process they check out what it's used for and make sure they're doing the right thing.

dbt_semantic_interfaces/naming/dundered.py
courtneyholcomb: Wait ignore me you're right. Minor is not the same is patch. Sorry my brain is busted rn 🙃

dbt_semantic_interfaces/parsing/where_filter/parameter_set_factory.py
WilliamDee: Ooo yea, ill drop the default

Issue #362, opened 2024-11-05
[Feature] Support for meta in Dimensions, Entities, and Measures


DevonFulcher:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward change to existing dbt-semantic-interfaces functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Dimensiona, Entities, and Measures should have meta config like other dbt resources https://docs.getdbt.com/reference/resource-configs/meta.

### Who will this benefit?

This will provide consistency across dbt resources.

### Are you interested in contributing this feature?

Yes.

### Anything else?

I have already started on this [here](https://github.com/dbt-labs/dbt-semantic-interfaces/pull/358). The biggest open question is how to implement combining configs, such as here https://docs.getdbt.com/reference/configs-and-properties#combining-configs. This behavior is a bit confusing to me and there may be a problem with our docs our a bug in the implementation. For example, In this case, there is no `meta_location_1` in the manifest:
```
sources:
  - name: source_1
    config:
      meta: {"meta_location": "source level1", "meta_location_1": "source level2"}

    tables:
      - name: table_1
        config:
          meta: {"meta_location": "table level3", "meta_location_2": "table level4"}
```
However, in this case there is:
```
sources:
  - name: source_1
    config:
      meta: {"meta_location": "source level1", "meta_location_1": "source level2"}

    tables:
      - name: table_1
```
So, it looks like these meta dictionaries aren't being merged as described in the docs.

Jstein77:
@DevonFulcher I think we can move forward with the behavior described in the docs and open a core issue to fix the "clobbering" behaviour. How does that sound?

PR #367, opened 2024-11-11
Merge semantic model meta config

DevonFulcher:
### Description

This PR implements the meta config merge behavior described [here](https://docs.getdbt.com/reference/configs-and-properties#combining-configs) and referenced [here](https://github.com/dbt-labs/dbt-semantic-interfaces/issues/362).

I decided to stack this PR on top to keep it distinct from the other PRs, but I don't intend to merge this one into main directly. I will merge it into its parent branch first.

### Checklist

- [ ] ~~I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)~~ I'm not going to merge this directly into main, so we don't need a separate changelog entry.


WilliamDee (approved):

dbeatty10 (commented):
@DevonFulcher I couldn't tell if the tests are verifying that `meta` is "merging" rather than "clobbering".

Is there a way to add to the YAML contents and assertions of test case so that we know that it is merging rather than clobbering? I'm thinking a couple small additions could help differentiate.

tests/parsing/test_semantic_model_parsing.py
```diff
@@ -538,3 +538,85 @@ def test_semantic_model_dimension_validity_params_parsing() -> None:
...
+    assert semantic_model.dimensions[0].config.meta["sm_metadata"] == "qwer"
+    assert semantic_model.dimensions[0].config.meta["dim_metadata"] == "fdsa"
+    assert semantic_model.dimensions[1].config is not None
+    assert semantic_model.dimensions[1].config.meta["sm_metadata"] == "zxcv"
+    assert semantic_model.dimensions[1].config.meta["dim_metadata"] == "mlkj"
+    assert semantic_model.dimensions[2].config is not None
+    assert semantic_model.dimensions[2].config.meta["sm_metadata"] == "asdf"
+    assert semantic_model.dimensions[3].config is not None
+    assert semantic_model.dimensions[3].config.meta["dim_metadata"] == "gfds"
+    assert semantic_model.dimensions[3].config.meta["sm_metadata"] == "asdf"
```
dbeatty10, line 616: These lines cover the test cases I was thinking about here: https://github.com/dbt-labs/dbt-semantic-interfaces/pull/367#pullrequestreview-2430471281 ✅ 

Namely, it verifies that it _merged_ the two dictionaries and has the `sm_metadata` key/value inherited from the semantic model as well as the `dim_metadata` key/value defined on the dimension.

dbeatty10: Per discussion with @DevonFulcher, we decided that enumerating similar test cases for entities and measures would be overkill because they are covered by the tests for dimensions.

The reason is that dimensions, entities, and measures are all treated the same here:

```python
                    elements: Sequence[Union[PydanticDimension, PydanticEntity, PydanticMeasure]] = [
                        *sm.dimensions,
                        *sm.entities,
                        *sm.measures,
```

dbeatty10 (approved):
🚀

PR #368, opened 2024-11-12
Remove unnecessary check against legacy field

WilliamDee:
### Description
We decided that `Metric.type_params.grain_to_date` will not be changed to support custom grain, but this validation makes it break since it expects it to be the same typing.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/metrics.py
+++ b/dbt_semantic_interfaces/validations/metrics.py
@@ -76,7 +76,8 @@ class CumulativeMetricRule(SemanticManifestValidationRule[SemanticManifestT], Ge
                     else None
                 )
                 if (
-                    type_params_field_value
+                    field == "window"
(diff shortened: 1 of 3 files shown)


dbt_semantic_interfaces/validations/metrics.py
```diff
@@ -76,7 +76,8 @@ def validate_manifest(semantic_manifest: SemanticManifestT) -> Sequence[Validati
                     else None
                 )
                 if (
-                    type_params_field_value
+                    field == "window"
```
courtneyholcomb, line 79: does this not impact grain to date ?

courtneyholcomb: or do we just not have a test for that?

courtneyholcomb (approved):

PR #371, opened 2024-12-03
Handle default of `None` properly for `cumulative_type_params`

courtneyholcomb:
dbt-core parses YAML differently than DSI. There, the YAML dict for a metric ends up with  if that key is empty. Then when we call type_params.get(cumulative_type_params, {}), the result is None because the key does exist in the dict. This results in a key error when we finally call .get(grain_to_date). Thus, we must set the default value outside of .get().

Resolves #

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/metric.py
+++ b/dbt_semantic_interfaces/implementations/metric.py
@@ -209,8 +209,8 @@ class PydanticMetric(HashableBaseModel, ModelWithMetadataParsing, ProtocolHint[M
         data = deepcopy(input)
 
         # Ensure grain_to_date is lowercased
-        type_params = data.get("type_params", {})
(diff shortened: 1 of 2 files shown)


WilliamDee (approved):

dbt_semantic_interfaces/implementations/metric.py
```diff
@@ -209,8 +209,8 @@ def parse_obj(cls, input: Any) -> PydanticMetric:
         data = deepcopy(input)
 
         # Ensure grain_to_date is lowercased
-        type_params = data.get("type_params", {})
-        grain_to_date = type_params.get("cumulative_type_params", {}).get("grain_to_date")
+        type_params = data.get("type_params") or {}
```
theyostalservice, line 212: brutal

theyostalservice (approved):

PR #374, opened 2025-02-10
feat: add CODEOWNERS file

Cr0n1c:
This PR adds a CODEOWNERS file to the repository.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -10,7 +10,7 @@
 
 # In each subsection folders are ordered first by depth, then alphabetically.
 # This should make it easy to add new rules without breaking existing ones.
-
+* @dbt-labs/semantic-layer
 
 # Parsing, validations, and protocols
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -10,7 +10,7 @@
 
 # In each subsection folders are ordered first by depth, then alphabetically.
 # This should make it easy to add new rules without breaking existing ones.
-
+* @dbt-labs/semantic-layer
 
 # Parsing, validations, and protocols
 /dbt_semantic_interfaces/**  @dbt-labs/semantic-layer @QMalcolm @graciegoheen
```
aiguofer, line 16: Should we remove the team tag from this line then?

aiguofer (approved):

.github/CODEOWNERS
Cr0n1c: We can, if you don't anticipate any other teams owning any portion than I will remove it... just lmk

PR #379, opened 2025-06-24
Transform `agg_time_dimension` to lowercase

courtneyholcomb:
### Description
Seeing an error in MetricFlow where we can't find the `agg_time_dimension` in the semantic model because it's uppercased in the `defaults` section but lowercased (by transformation, presumably) in the `dimensions` section. This will ensure both are lowercased.

### Checklist


QMalcolm (approved):
I see no problem with this. As a heads-up, on the dbt-core side we don't run the transformation rules. So for this to take effect, this transformation will have to be run on the MetricFlow side

QMalcolm:
And it does look like the rule is being run [here in MetricFlow](https://github.com/dbt-labs/metricflow/blob/1637bf53f467a29734428c0051eaba789c6cc680/metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py#L32), so all should be good 🙂

PR #390, opened 2025-08-19
Add measures fields to metrics classes

theyostalservice:
Resolves #

### Checklist


theyostalservice:
* **#390** <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-semantic-interfaces/390?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-semantic-interfaces/390?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

dbt_semantic_interfaces/implementations/metric.py
```diff
@@ -168,6 +173,18 @@ class PydanticCumulativeTypeParams(HashableBaseModel):
     period_agg: PeriodAggregation = PeriodAggregation.FIRST
 
 
+class PydanticMetricAsSourceNodeParams(HashableBaseModel):
```
courtneyholcomb, line 176: Alright let's talk naming! I feel like `MetricAsSourceNode` is confusing since we don't really use the terms "source" or "node" in this context. We do in MetricFlow, but there, a "source node" refers to the semantic model. The change here at a high level is that we're applying the aggregation at the metric level, instead of the measure level. So I'm thinking the name should have something to do with metric aggregation. I like `PydanticMetricAggregationParams` or something similar to that. (Same for the protocol)

courtneyholcomb: Noticing that this proposed name is weird when you see the class contains `PydanticMeasureAggregationParameters`... but I feel like that one was a bad name? Would love to change it to be percentile-specific since that's all it contains but I don't think it's worth dealing with the downstream implications of a breaking change here. If you hate those two names next to each other we could potentially recreate that class with a new name and use that here, but I'll leave it up to you on that front.

courtneyholcomb (approved):
Everything looks good except for some workshopping on the name!

PR #406, opened 2025-09-02
Remove non-working validation

courtneyholcomb:
### Description
This validation was never actually working due to contradictory lines of code. Removing it so that no one makes the false assumption that dimensions with the same element name must have the same grain.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/validations/dimension_const.py
+++ b/dbt_semantic_interfaces/validations/dimension_const.py
@@ -1,15 +1,10 @@
 from typing import Dict, Generic, List, Sequence
 
-from dbt_semantic_interfaces.protocols import (
-    Dimension,
-    SemanticManifestT,
-    SemanticModel,
-)
+from dbt_semantic_interfaces.protocols import SemanticManifestT, SemanticModel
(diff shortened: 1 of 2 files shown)


dbt_semantic_interfaces/validations/dimension_const.py
```diff
@@ -33,67 +28,13 @@ class DimensionConsistencyRule(SemanticManifestValidationRule[SemanticManifestT]
...
-            element_type=SemanticModelElementType.DIMENSION,
-        )
-
-        if dimension.type == DimensionType.TIME:
-            if dimension.reference not in time_dims_to_granularity and dimension.type_params:
-                time_dims_to_granularity[dimension.reference] = dimension.type_params.time_granularity
-
-                if (
-                    dimension.type_params is not None
-                    and dimension.type_params.time_granularity != time_dims_to_granularity[dimension.reference]
```
courtneyholcomb, line 85: Line 81 above guarantees that this line will never be truthy, so this validation will never run.

WilliamDee (approved):

Issue #387, opened 2025-07-30
[Task] Update YAML spec to better unify with dbt and fusion for metrics and measures


theyostalservice:
### Is this your first time submitting a feature request?

- [x] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [x] I have searched the existing issues, and I could not find an existing issue for this feature
- [x] I am requesting a straightforward change to existing dbt-semantic-interfaces functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Unnecessary variance in the way we define metrics and measures in MF and other dbt work make it harder for folks using one to understand and use the other.

We've identified some incremental improvements in the YAML spec that should make it more flexible and able to support a more similar workflow.

### Who will this benefit?

Everyone who would like to use semantic layer and another dbt product.

### Are you interested in contributing this feature?

Yes

PR #436, opened 2025-10-01
Add measures to metrics (temporary) & fix metric deduplication

theyostalservice:
Towards #387

### Description

This PR contains a few small changes and improvements, all grouped up in the process of getting to a point where metricflow can onboard onto the new dsi version immediately.

At a high level, these changes take three forms.  I've tried to detail in comments throughout the PR, but here as well:
* general clean up or improvements that came up along the way
* fixing the equality checking when searching for matching metrics (to avoid creating duplicates).
  * this is a correctness issue here in this repo.  It can arise when we have a metric created from a measure with create_metric=true and then also create one for inputs.  (This was found in some of the "parse-the-entire-project" tests in metricflow.)
* adding `measure`s and sometimes `input_measures` to some of the new metrics made in transformations.  This is only necessary because the upgrades in MF are not yet complete and some operations still throw errors when encountering metrics without input measures (especially in tests :/ ), and these new lines can be removed in a few weeks when everything is ready.

I've also tested this by pointing the metricflow repo at the top commit here to make sure `make test` is running alright since the latest dsi version would not work in MF before.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/implementations/metric.py
+++ b/dbt_semantic_interfaces/implementations/metric.py
@@ -347,11 +347,14 @@ class PydanticMetric(HashableBaseModel, ModelWithMetadataParsing, ProtocolHint[M
 
         It lives here instead of measures to avoid circular import issues.
         """
+        agg_params = measure.agg_params.copy(deep=True) if measure.agg_params is not None else None
+        non_additive_dimension = (
+            measure.non_additive_dimension.copy(deep=True) if measure.non_additive_dimension is not None else None
+        )
         return PydanticMetricAggregationParams(
             semantic_model=semantic_model_name,
             agg=measure.agg,
-            # TODO: clone this
-            agg_params=measure.agg_params,
+            agg_params=agg_params,
             agg_time_dimension=measure.agg_time_dimension,
-            non_additive_dimension=measure.non_additive_dimension,
+            non_additive_dimension=non_additive_dimension,
         )
--- a/dbt_semantic_interfaces/implementations/metric.py
+++ b/dbt_semantic_interfaces/implementations/metric.py
@@ -350,6 +350,7 @@ class PydanticMetric(HashableBaseModel, ModelWithMetadataParsing, ProtocolHint[M
         return PydanticMetricAggregationParams(
             semantic_model=semantic_model_name,
             agg=measure.agg,
+            # TODO: clone this
             agg_params=measure.agg_params,
             agg_time_dimension=measure.agg_time_dimension,
             non_additive_dimension=measure.non_additive_dimension,
--- a/dbt_semantic_interfaces/transformations/add_input_metric_measures.py
+++ b/dbt_semantic_interfaces/transformations/add_input_metric_measures.py
@@ -31,9 +31,7 @@ class AddInputMetricMeasuresRule(ProtocolHint[SemanticManifestTransformRule[Pyda
     ) -> Set[PydanticMetricInputMeasure]:
         """Returns a unique set of input measures for a given metric."""
         measures: Set = set()
-        matched_metric = next(
-            iter((metric for metric in semantic_manifest.metrics if metric.name == metric_name)), None
-        )
+        matched_metric = next((metric for metric in semantic_manifest.metrics if metric.name == metric_name), None)
         if matched_metric:
             if matched_metric.type is MetricType.SIMPLE or matched_metric.type is MetricType.CUMULATIVE:
                 if matched_metric.type_params.measure is not None:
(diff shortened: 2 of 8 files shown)


tests/parsing/test_metric_parsing_with_custom_grain.py
```diff
@@ -48,6 +48,7 @@ def test_cumulative_metric_with_custom_grain_to_date() -> None:  # noqa: D
           type_params:
             measure:
               name: bookings
+              fill_nulls_with: 15
```
theyostalservice, line 51: de-duplicating cut the number of metrics down by 1, but I wanted to make sure we had variety in our tests so I made this one require a new metric.

dbt_semantic_interfaces/transformations/measure_to_metric_transformation_pieces/measure_features_to_metric_name.py
```diff
@@ -83,7 +99,7 @@ def build_metric_from_measure_configuration(
         semantic_model_name: str,
         fill_nulls_with: Optional[int],
         join_to_timespine: Optional[bool],
-        is_private: bool = True,
+        is_private: bool,
```
theyostalservice, line 102: this is just an improvement - the default value here was just making the code less readable elsewhere.

dbt_semantic_interfaces/implementations/metric.py
```diff
@@ -347,10 +347,14 @@ def build_metric_aggregation_params(
 
         It lives here instead of measures to avoid circular import issues.
         """
+        agg_params = measure.agg_params.copy(deep=True) if measure.agg_params is not None else None
```
theyostalservice, line 350: this isn't necessary for the fix, but it makes things better/safer in the future.

dbt_semantic_interfaces/transformations/add_input_metric_measures.py
```diff
@@ -31,9 +31,7 @@ def _get_measures_for_metric(
     ) -> Set[PydanticMetricInputMeasure]:
         """Returns a unique set of input measures for a given metric."""
         measures: Set = set()
-        matched_metric = next(
-            iter((metric for metric in semantic_manifest.metrics if metric.name == metric_name)), None
-        )
+        matched_metric = next((metric for metric in semantic_manifest.metrics if metric.name == metric_name), None)
```
theyostalservice, line 34: (also just clean up)

dbt_semantic_interfaces/transformations/measure_to_metric_transformation_pieces/measure_features_to_metric_name.py
```diff
@@ -65,16 +66,31 @@ def _find_metric_clone_in_manifest(
...
         search_metric = metric.copy(deep=True)
+        original_input_measures = search_metric.type_params.input_measures
+        original_measure = search_metric.type_params.measure
         for existing_metric in manifest.metrics:
             # this allows us to a straight equality comparison, which is safer in the future
             # than implementing a custom comparison function.
             search_metric.name = existing_metric.name
+            search_metric.description = existing_metric.description
+            search_metric.label = existing_metric.label
             search_metric.metadata = existing_metric.metadata
```
theyostalservice, line 77: we don't care about these fields either when we search for a match!

courtneyholcomb (commented):
One main concern!

dbt_semantic_interfaces/implementations/metric.py
theyostalservice: It's mostly a safety decision / data hygiene.  Since these are objects, I'd rather not assign them directly so that a future transformation to a metric won't change the original measure's agg_params or vice versa, for example.

dbt_semantic_interfaces/transformations/measure_to_metric_transformation_pieces/measure_features_to_metric_name.py
```diff
@@ -65,16 +66,31 @@ def _find_metric_clone_in_manifest(
...
+            if len(existing_metric.type_params.input_measures) == 0:
+                search_metric.type_params.input_measures = []
+            else:
+                search_metric.type_params.input_measures = original_input_measures
+            if existing_metric.type_params.measure is None:
+                search_metric.type_params.measure = None
+            else:
+                search_metric.type_params.measure = original_measure
+
             if search_metric == existing_metric:
```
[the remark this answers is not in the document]
theyostalservice: When it was just a few fields, I felt a lot more strongly about trying this strategy, but it's ballooned enough that I can see the advantage of trading it back to using individual fields.

If we want this to be generalizable, I don't think changing is more future proof (any new field would need to be accounted for, and that's why you'd want to avoid enumerating them), but I can add a note to the doc limiting the usecases for this function.

courtneyholcomb (approved):
lgtm!

theyostalservice:
### Merge activity

* **Oct 3, 12:15 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dbt-labs/dbt-semantic-interfaces/436).

PR #439, opened 2025-10-07
Copy FixSimpleMetricRule into dsi

theyostalservice:
Towards #387

### Description

Paul ran into some issues in MF with some of the newly-minted simple metrics from transformations.  Ideally, it might arguably be better to fold the necessary improvements into code in the existing transformations, but he has a working transformation to fix them all and I don't have a good repro test case, so we're just going to add another rule to clean them all up.

I split the commit where I copied the file in apart from the commits where I made actual changes, so review-by-commit should make this a really fast and easy review.

### Checklist

### DIFF
--- a/dbt_semantic_interfaces/transformations/fix_simple_metric_rule.py
+++ b/dbt_semantic_interfaces/transformations/fix_simple_metric_rule.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 import copy
 import logging
 from collections.abc import Mapping
-from typing import override
+
+from typing_extensions import override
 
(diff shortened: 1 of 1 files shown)


courtneyholcomb (commented):
This doesn't feel like a great thing to put in open source. I think this was an oversight in the other transformations, and what we actually want is to update those transformations to do this work, right?
If you don't think this logic fits into other transformations we can at least rename this to sound less hacky.
Also, can we add tests for the cases that this handles?

If you're concerned that changing this will take too long, LMK!

PR #461, opened 2026-01-22
Fix FixProxyMetricsRule transformation to properly log the warning

WilliamDee:
### Description
Currently, we are throwing this log whenever the `metric.type_params.expr` is populated cause we assume that if it is populated it is potentially incorrect. This is true, however, we actually ran the `CreateProxyMeasureRule` prior to this, so all the proxy metrics would already have the expr populated in this case. It's a matter of whether it was populated correctly, so instead we would check that it should be matching the measure's designated expr.

### Checklist


courtneyholcomb (approved):
thank you!

PR #469, opened 2026-03-26
Add compiled_sql to NodeRelation for ephemeral model support

b-per:
## Summary

- Adds optional `compiled_sql` field to the `NodeRelation` protocol and `PydanticNodeRelation` implementation
- Updates JSON schema and `schemas.py` to allow `compiled_sql` in YAML parsing
- When set, `compiled_sql` carries the compiled SQL for ephemeral dbt models instead of relying on `relation_name`

This is the upstream companion to dbt-labs/metricflow#1993, which already consumes this field via `getattr` as a workaround until this PR lands. Once merged and published, the `getattr` workaround in metricflow can be replaced with a direct attribute access.

## Test plan

- [x] `tests/implementations/test_node_relation.py` — new unit tests for `compiled_sql` default, setting, and `from_string` behavior
- [x] `tests/parsing/test_semantic_model_parsing.py::test_semantic_model_node_relation_with_compiled_sql` — new parsing test for ephemeral model YAML
- [x] `tests/parsing/test_semantic_model_parsing.py::test_semantic_model_node_relation_parsing` — updated to also assert `compiled_sql is None` for non-ephemeral models

### DIFF
--- a/.changes/unreleased/Features-20260326-000000.yaml
+++ b/.changes/unreleased/Features-20260326-000000.yaml
@@ -1,0 +1,5 @@
+kind: Features
+body: Add compiled_sql to NodeRelation protocol and PydanticNodeRelation to support ephemeral models in Semantic Layer definitions.
+time: 2026-03-26T00:00:00.000000-07:00
+custom:
+  Author: b-per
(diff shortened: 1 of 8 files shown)


b-per:
CI failures are unrelated to this change — all jobs fail with `module 'virtualenv.discovery.builtin' has no attribute 'propose_interpreters'` before any test code runs. Looks like a virtualenv version bump broke hatch in the CI environment.

PR #474, opened 2026-09-11
Skip measure alias validation for derived and ratio metrics (#141)

### LINKED ISSUE #141: [Bug] Derived metric validation false positive on "identical input measures"

### Current Behavior

Running validation on a model with a metric specified like so:

```
      metrics:
        - name: revenue
          alias: current_revenue
        - name: revenue
          offset_window: 1 month
          alias: revenue_prev_month
```

Prints this warning:

```
15:07:37  Running with dbt=1.6.1
15:07:37  Registered adapter: snowflake=1.6.2
15:07:38  PydanticMetric revenue_growth_mom has multiple identical input measures specifications for measure revenue. This might be hiding a semantic error. Input measure specification: name='revenue' filter=None alias=None.
```

The issue here is we are running the input measures validator on derived metrics without accounting for the fact that a derived metric is made up of multiple metrics which might themselves be sourced from shared measures.

There's nothing a user can do to fix this or act upon it. Indeed, nothing is even broken - in this case the user wants to compute some expr based on a transformation of the same input metric. That transformation and alias will naturally apply to the measure input on the metrics, but we can't detect that at this level, so we raise the validation warning about shared input measures (duh) with a baffling "alias: None" in there (because the input measures don't have aliases set).

### Expected Behavior

We would expect this to pass validation cleanly, as nothing is wrong.

More generally, derived metric input validation should not be validating input measures. It should be validating input metrics.

### Steps To Reproduce

Define a derived metric that relies on two distinct transformations of the same input metric (offset or filter) with aliases properly set and run the validation in a mode that will print warnings.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- dbt-semantic-interfaces:
```

### Additional Context

_No response_

cpruijsen:
Resolves #141

### Description

`MeasureConstraintAliasesRule` ran on every metric, including derived and ratio, using flattened child `input_measures`. Those rows drop metric-input offset/alias/filter, so a derived metric built from two simple metrics that share a measure looked like unaliased constrained measures. Users cannot set those aliases on the derived/ratio YAML; the rule raised `aliases must be provided`. The original #141 warning (`multiple identical input measures specifications` with `alias=None`) is the same rule on identical flattened rows.

#141: derived metric input validation should not be validating input measures; it should be validating input metrics. `DerivedMetricRule` already validates input metrics. This PR skips the measure-alias rule for `DERIVED` and `RATIO` only.

**What I chose and the alternative:** skip both derived and ratio. Alternative: skip derived only, or add new alias-required checks on input metrics. Both types compose input metrics the same way (`AddInputMetricMeasuresRule`); new alias requirements would be extra behaviour. Happy to drop the ratio skip if that is preferred.

Simple, cumulative, and conversion metrics are unchanged.

Note: on current main, `AddInputMetricMeasuresRule` stores a set of child measures, so the exact MoM YAML from the issue (same input metric twice with offsets and aliases) no longer emits the warning. Two simple metrics sharing a measure with different filters still error without this skip.

### Checklist


cpruijsen:
> Thanks for your pull request, and welcome to our community! We require contributors to sign our [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScfOV7K4enYRHozrDRP6BBIXjOij-JDGca6WBTHyP_ANXSqlg/viewform) and we don't seem to have your signature on file. Check out [this article](https://docs.getdbt.com/docs/contributor-license-agreements) for more information on why we have a CLA.
> 
> In order for us to review and merge your code, please submit the Individual Contributor License Agreement form attached above above. If you have questions about the CLA, or if you believe you've received this message in error, please reach out through a comment on this PR.
> 
> CLA has not been signed by users: @cpruijsen

Signed now

cpruijsen:
I think the CLA bot is stuck - I did sign the Google form

cpruijsen:
@cla-bot check

